PR #28, opened 2016-07-01
Fix charset to fit IANA standard

AArnott:
Per the character sets listed in the standard: http://www.iana.org/assignments/character-sets/character-sets.xhtml, 'utf8' is not the name of the character encoding. The correct encoding name is 'utf-8'. Correcting this brings it in conformance with web standards, allowing more interoperability. In particular .NET's Encoding.GetEncoding(string) function only returns an Encoding instance for 'utf-8' and null for 'utf8'.

See also: http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7

Note that correcting the doc is just one step. The code that produces and consumes this header in VS Code should also be checked and fixed if necessary.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -29,7 +29,7 @@ Currently the following header fields are supported:
 | Header File Name | Value Type  | Description |
 |:-----------------|:------------|:------------|
 | Content-Length   | number      | The length of the content part in bytes. This header is required. |
-| Content-Type     | string      | The mime type of the content part. Defaults to application/vscode-jsonrpc; charset=utf8 |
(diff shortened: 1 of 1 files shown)


dbaeumer:
@AArnott makes perfect sense to me. What I propose is to merge that into a 3.0 version of the protocol since as you noticed it is breaking.

dbaeumer:
I decided to change this is the spec since the right encoding for other languages is 'utf-8' and it is causing more confusion then it helps keeping this compatible (done via another issue). The VS Code node modules are adopted to be able to handle both 'utf8' and 'utf-8'

dbaeumer:
Just to clarify: I updated the spec accordingly.

PR #36, opened 2016-07-09
DidOpenTextDocumentParams does not extend TextDocumentIdentifier

Marwes:
See https://github.com/Microsoft/vscode-languageserver-node/blob/032c12a09cf19ad7fb6af997beca1d5f6a7f34b2/server/src/protocol.ts#L411-L416.

The `uri` field is present in the `TextDocumentItem` in any case.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -765,7 +765,7 @@ _Notification_:
 * method: 'textDocument/didOpen'
 * params: `DidOpenTextDocumentParams` defined as follows:
 ```typescript
-interface DidOpenTextDocumentParams extends TextDocumentIdentifier {
+interface DidOpenTextDocumentParams {
 	/**
 	 * The document that was opened.
 	 */


Marwes:
Sorry, now you got me confused. Is DidOpenTextDocumentParams supposed to extend TextDocumentIdentifier (which gives it an `uri` field directly in addition to the one in `textDocument`) but that is not implemented yet in https://github.com/Microsoft/vscode-languageserver-node ?

dbaeumer:
@Marwes no, it was that way in the 1.0 version of the protocol (which is available here: https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-1-x.md). In the 2.0 version we cleaned this up and steps away from 'inlining' parameters.

Marwes:
In that case, should my PR not be a fix since it removes ` extends TextDocumentIdentifier` ?

dbaeumer:
@Marwes I apologoize a lot. I misread the diff. Makes perfect sense. And thank you for not giving up :-).

Marwes:
No problem. Just happy to be of some help, the protocol made it really simple to provide completion for my own little language :).

PR #40, opened 2016-07-22
json schema for protocol

gorkem:
A Json schema that is mostly generated from protocol.d.json

### DIFF
--- a/protocol.json
+++ b/protocol.json
@@ -1,0 +1,701 @@
+{
+    "type": "object",
+    "properties": {
+        "range": {
+            "$ref": "#/definitions/Range"
+        },
+        "position": {
+            "$ref": "#/definitions/Position"
+        },
+        "location": {
+            "$ref": "#/definitions/Location"
+        },
+        "diagnostic": {
+            "$ref": "#/definitions/Diagnostic"
+        },
+        "DiagnosticSeverity": {
+            "$ref": "#/definitions/DiagnosticSeverity"
+        },
+        "command": {
+            "$ref": "#/definitions/Command"
+        },
+        "textEdit": {
+            "$ref": "#/definitions/TextEdit"
+        },
+        "workspaceEdit": {
+            "$ref": "#/definitions/WorkspaceEdit"
+        },
+        "workspaceChange": {
+            "$ref": "#/definitions/WorkspaceChange"
+        },
+        "textEditChange": {
+            "$ref": "#/definitions/TextEditChange"
+        },
+        "textDocumentIdentifier": {
+            "$ref": "#/definitions/TextDocumentIdentifier"
+        },
+        "completionItemKind": {
+            "$ref": "#/definitions/CompletionItemKind"
+        },
+        "completionItem": {
+            "$ref": "#/definitions/CompletionItem"
+        },
+        "completionList": {
+            "$ref": "#/definitions/CompletionList"
+        },
+        "hover": {
+            "$ref": "#/definitions/Hover"
+        },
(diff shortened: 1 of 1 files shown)


nojvek:
Pretty awesome that this confirms to the json schema definition. 

Could you link the Java generator ?

gorkem:
the pom.xml that generates java model from json schema https://github.com/gorkem/java-language-server/blob/master/server/lang-protocol-generator/pom.xml

dbaeumer:
@gorkem: nice work. Thanks a lot. Before we merge this in I have a couple of questions / suggestions:

- do you know of any tool to create a ts file from a schema.
- and shouldn't we group the json schema a little bit. What I mean is that everything for a request is at one place?
- and it looks like enums are translated wrongly if they have assigned numbers. An example is MessageType which is in the schema:
```json
"MessageType": { 
    "type": "string" 
}, 
```
but in TS 
```ts
/**
 * The message type
 */
export enum MessageType {
	/**
	 * An error message.
	 */
	Error = 1,
	/**
	 * A warning message.
	 */
	Warning = 2,
	/**
	 * An information message.
	 */
	Info = 3,
	/**
	 * A log message.
	 */
	Log = 4
}
```
- looks like we lost comments as well (especially all the interface comments)
- Request and Notification type definitions aren't included either. Example is: ShutdownRequest

Did you generate the file with a tool. May be we can tweak it a little for our purpose.

dbaeumer:
@gorkem and you need to sign the CLA (see https://github.com/Microsoft/language-server-protocol/pull/40#issuecomment-234588746). Otherwise I can't merge it in.

svenefftinge:
@gorkem are you going to sign the CLA and work on the things @dbaeumer mentioned? We would like to use this as the source for our code generator in https://github.com/TypeFox/ls-api

gorkem:
@svenefftinge Unfortunately, we needed a CCLA so this is waiting on that.

felixfbecker:
@dbaeumer I only know about the other way: https://github.com/YousefED/typescript-json-schema

protocol.json
```diff
@@ -0,0 +1,1216 @@
+{
+    "type": "object",
```
felixfbecker: Should this really be an object? Shouldn't the schema only define all structures?

gorkem: It is something that the generator tool created. We should remove it going forward.

gorkem:
@dbaeumer I should have a CLA now, is there a way to check it has gone in effect?

dbaeumer:
@gorkem usually the cla label is updated by the cla bot. Let me see what is missing here.

dbaeumer:
Looks like closing and reopening should trigger a revalidate. Let me see if it works.

dbaeumer:
OK. Here we go.

Instead of simply applying this PR I think we also need to work on a tool chain to generate the schema from the protocol.ts file here: https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/protocol.ts

Maintaining the schema by hand will be a nightmare and I don't want to make json schema the master :-)

@gorkem which tool did you use to generate the JSON schema.

gorkem:
I have used [typescript-json-schema ](https://www.npmjs.com/package/typescript-json-schema). It needs to be massaged in some cases but works most of the time. I also have a Gruntfile I used that automates things which I can share.

dbaeumer:
Great. Can you attach it here?

dbaeumer:
I will close the PR. If we want a json schema we should invest in some automation.

PR #46, opened 2016-07-29
Consistent use of suffix 'Request' in headers

kthoms:
In previous sections, the suffix 'Notification' and 'Request' was added as suffix in the section header. Added the suffix 'Request' for the section headers where this was missing.


dbaeumer:
Makes sense.

dbaeumer:
and thanks!

PR #85, opened 2016-10-05
Fix typo in protocol documentation

masaeedu:


ags799:
Would appreciate confirmation that the intended text is `currently no` and not `currently now`.

masaeedu:
The sentence "there is currently now support" is grammatically incorrect, and would also contradict the previous sentence. The current state of the protocol does not support sharing a single language server between different tools.

dbaeumer:
Thanks for the PR

Issue #72, opened 2016-09-20
Clarification on MarkedString


gorkem:
I think the role of language needs to be clarified on MarkedString. It is not obvious from the documentation what it is used for.

dbaeumer:
Improved doc around this.

Issue #78, opened 2016-09-30
Clarify meaning of range within Hover


Et999:
It is not clear what it should be the range of.

akosyakov:
It's a range at which hover should be used. I guess a client side should not send a new hover request till the user hovers over the range of the current range. If the range is undefined then the client should send a new request on a cursor position change.

ljw1004:
Some clients highlight the symbol under the cursor when you hover -- so the user clearly understands "this hover tooltip applies to this precise symbol". This range gives them the opportunity to know which range to highlight.

PR #91, opened 2016-10-09
Fixes issues #72 and #78

akosyakov:


protocol.md
```diff
@@ -1078,13 +1078,24 @@ interface Hover {
...
+	 * that is used to visualize a hover, e.g. by changing the background color.
 	 */
 	range?: Range;
 }
 ```
 Where `MarkedString` is defined as follows:
 ```typescript
+/**
+ * The marked string is rendered:
+ * - as markdown if it is represented as a string
```
mickaelistria: Not sure this is something the server should specify.

akosyakov: Which part? It does not expose anything about the markdown format, so it is still up to the server to decide what kind of markdown to send. I've just wanted to point out that there is semantical difference depending on the type of a string. We've already had troubles to decided what to send.

mickaelistria: Specifying default is markdown is IMO not making the protocol better.
If your patch is amending the current version of the protocol, then it should say that the marked string is *usually* rendered as markdown since it may not be the choice taken by all clients.
If your patch is a proposal for a future version of the protocol, then simply finding a way to avoid ambiguity (remove string and keep only MarkedString) seems much better.

akosyakov: I just state how they are handled now. Agreed about `usually` for markdown.

I don't see why `MarkedString` is better than `string`? I would prefer strings, since they are generic and can encode code blocks. It would simplify specification to a simple fact that `marked string are usually rendered as markdown`. Opposite is not true, one need to come up with a special handling of a language to allow arbitrary markdown.

mickaelistria: Why forcing usage of markdown in the protocol when some other formats exists? Some servers may prefer HTML or asciidoc. Giving the choice to the server is actually a good thing, and if we want to give choice, then the MarkedString needs to remain to explicit the language; and if we have to keep MarkedString for that, why also keeping string?

svenefftinge: giving choice to the server comes at the cost that clients need to support the different formats.

dbaeumer: Actually the idea was to use markdown :-) As the content definition says it is a `MarkedString | MarkedString[]`. So I will merge the PR in to make this more clear since it is todays state. I opened https://github.com/Microsoft/language-server-protocol/issues/94 to continue the discussion whether this was a good idea or not.

PR #97, opened 2016-10-19
Link directly to vscode blogpost

gmorenz:
Change link from https://code.visualstudio.com/blogs/ to https://code.visualstudio.com/blogs/2016/06/27/common-language-protocol


gregvanl:
Thanks!

Issue #13, opened 2016-05-31
Support late resolution of symbol (reference) information


akosyakov:
It would be nice if a location and a container of symbol are resolved later. In our case we have an index of symbols but it is missing information about locations, so to compute them we have to load models for all found symbols. I assume that locations are required for navigation and their can be computed on demand later. The same should be true for references.

dbaeumer:
Good point. This is currently possible for completion items and code actions but not for symbol references.

felixfbecker:
Yes please!

yaohaizh:
Any update on this feature request?  The WorkspaceSymbolProvider.resolveWorkspaceSymbol have this feature but not on the language protocol side.

felixfbecker:
@yaohaizh this is also possible in a push manner instead of pull with #182

yaohaizh:
@felixfbecker, thanks. Will look into the streaming methods.

felixfbecker:
@yaohaizh let me know what you think! you can find a list of language servers that support streaming on http://langserver.org/

dbaeumer:
Implemented for workspace symbols for 3.17.0

dbaeumer:
Reopening, since it is still not there for references.

PR #101, opened 2016-10-20
Protocol addition to get SymbolInformation for a given text document position

uforic:
I'm developing a plugin to allow developers to look up global usages of a given symbol. I’ve also see plugins that automatically link you to the godoc for a symbol (in the case of Go), and plugins that automatically search StackOverflow or Google, to which this protocol change would also apply.

To build these plugins, one needs to know the information about the symbol under the cursor - for instance, in a go example, if the user wants to find the godoc page for `mux.Router`, we need to construct the url `https://godoc.org/github.com/gorilla/mux#Router`, which requires knowing  the name (Router), the parent, the package (mux), and the git url (github.com/gorilla/mux). The `SymbolInformation` data type returned by `workspace/symbol` and `textDocument/documentSymbol` is exactly what we need.

Currently, to create such a plugin and get the `SymbolInformation` required, I find myself needing to request `textDocument/definition` for the `TextDocumentPostionParams` in question. Then, I need to request `textDocument/documentSymbol,` and iterate through the returned symbols to find the one with the same `Location` as the `textDocument/definition.`

The problem with this approach is that it requires two requests, the second of which can be expensive. I wanted to propose a modification to the protocol that would solve the above problem, which seems like it will come up for many types of plugins one could imagine writing. Here are some options I considered:
1. `textDocument/definition` should return `SymbolInformation` (a superset of `Location`). Upside: Only one request to perform. Downside: Perhaps there are some implementations of `textDocument/definition` that are cheaper and faster without needing the full `SymbolInformation,` so we shouldn’t penalize this API call.
2. (The option presented in this PR) Given the downside from 1, creating a new protocol message `textDocument/symbol`, which returns `SymbolInformation`. Upside: Same as above. Downside: One more request to implement for language server creators.
3. Adding an optional `Location` parameter to the `textDocument/documentSymbol` request, that would focus the search in a text document to just the range attached to the location. Upside: Potentially can speed up `textDocument/documentSymbol` request by not analyzing and returning `SymbolInformation` for out of scope symbols. Downside: The above process still requires two requests, and performance difference will probably vary based on language server implementation.

Open to any other suggestions and feedback on how to solve the above problem, and curious to hear if anyone else is running into this.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -581,6 +581,10 @@ interface ServerCapabilities {
 	 */
 	definitionProvider?: boolean;
 	/**
+	 * The server provides get symbol support.
+	 */
+	symbolProvider?: boolean;
+	/**
 	 * The server provides find references support.
 	 */
 	referencesProvider?: boolean;
@@ -1207,6 +1211,18 @@ _Request_
 
 _Response_:
 * result: [`Location`](#location) | [`Location`](#location)[]
+* error: code and message set in case an exception happens during the definition request.
+
+#### Get Symbol Request
+
+The get symbol request is sent from the client to the server to resolve the definition location and symbol information for a given text document position.
+
+_Request_
+* method: 'textDocument/symbol'
+* params: [`TextDocumentPositionParams`](#textdocumentpositionparams)
+
+_Response_:
+* result: `SymbolInformation` | `SymbolInformation`[]
 * error: code and message set in case an exception happens during the definition request.
 
 #### Find References Request


akosyakov:
@uforic I will go for the first option, if the server knows a location then it should not be expensive to get an URI, a symbol name and a container name, a server specific additional symbol information can be resolved with another request: https://github.com/Microsoft/language-server-protocol/issues/13.

uforic:
Ah @sqs , I realize now I may have misread the document/symbol API call, `list all symbols found in a given text document.`

I assumed that meant all symbol definitions, not all symbols including references. If it's the latter, you're right, option 3 seems like the best way to go.

How do you interpret that text? I'm leaning towards your interpretation now.

felixfbecker:
I interpret `document/symbol` as list all symbol _declarations_. That is also how the TS language service behaves when you execute the command in VS Code.

sqs:
Indeed. I was wrong.

dbaeumer:
This makes sense but we shouldn't tweak an existing request for this. Thinking about it from a compiler language service point of we you want to resolve a symbol under a cursor position. So I would call the request `textDocument/resolveSymbol` and it should return a `ResolvedSymbolInformation`. The current SymbolInformation is mainly made for presenting Symbols in the user interface. However if you really resolve a symbol then for example you want to be able to navigate the parent chain as well (for example you have resolved a method and want to reach to the class containing the method).

uforic:
Solid, will open a new PR when I've had a chance to incorporate the feedback here. Thanks folks for the feedback.

vimpostor:
It's been now almost 10 years and as far as I can tell there is still no way to get information just for the symbol under the cursor.
Is iterating through all `textDocument/documentSymbol` results and matching them with the result of `textDocument/definition` really still the best we can do here (which is really flaky), or did we ever see an update on how to handle this usecase better?

My usecase is also showing documentation with the help of another non-LSP based external tool such as [Zeal](https://zealdocs.org/), but for doing so I need to know the complete symbol identifier for the symbol under the cursor.
A similar issue is also: https://github.com/microsoft/language-server-protocol/issues/1034

Issue #25, opened 2016-06-28
language server serves /protocol.json


nojvek:
It seems this protocol has similarities to the Chrome Remote Debugging protocol. 

Does it make sense for the repo to have a protocol.json file ? e.g https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/v8_inspector/js_protocol.json

This means different language servers implement the protocol and maintain a protocol.json file which can be queried if the API is compatible and has any breaking changes.

The protocol.json can then be used to automatically generate code for a client or server implementation in languages other than typescript.

e.g I can consume protocol.json and automatically generate a Java class for java language service interface.

Same for PHP, python and the rest. Code generation on both client and server can benefit.

gorkem:
+1 excellent idea! [Eclipse JSDT](http://git.eclipse.org/c/jsdt/webtools.jsdt.git/tree/bundles) uses protocol.json to generate the Java model.

dbaeumer:
Yes that makes sense to me. We could then even generate the protocol.ts from it.

gorkem:
I think I can help with this since I am creating a similar think for the java binding. What should be the meta model though. Use the model from V8s protocol.json or alternately use json schema specification?

nojvek:
I've made some progress on the last few days defining protocol.d.ts
interface for chrome's protocol.json files. Thanks to typescript I already
found a few nitty gritty bugs in their json files.

Hope it can be used as a template.

https://github.com/nojvek/chrome-remote-debug-interface/tree/master/generator

gorkem:
I created a [PR with the protocol.json ](https://github.com/Microsoft/language-server-protocol/pull/40) that I have used to generate the java-language-server APIs. The files is a json schema and it is mostly generated from protocol.d.ts file.

dbaeumer:
I will close this issue in favor of #67

PR #117, opened 2016-11-07
add table of contents to the specification #114

### LINKED ISSUE #114: Add table of contents to the specification
Hi!

Currently, the spec is a bit hard to read because there is no ToC which lists all methods/notifications and types. It might be a good idea to provide one! Take a look at the index at the bottom of the Dart server spec, for example, it is really convenient: https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#Index.

vladdu:
resolves #114

The TOC also has icons added to specify if the messages are requests or notifications and in which directions they go. These could also be added to the descriptions.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -10,6 +10,51 @@ This document describes version 2.x of the language server protocol. Changes are
 
 The 1.x version of this document can be found [here](https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-1-x.md).
 
+## Messages overview
+
+General
+
+* :leftwards_arrow_with_hook: [initialize](#initialize) 
+* :leftwards_arrow_with_hook: [shutdown](#shutdown) 
+* :arrow_right: [exit](#exit) 
+* :arrow_right: [$/cancelRequest](#cancelRequest)
+
+Window
+
+* :arrow_left: [window/showMessage](#window_showMessage) 
+* :arrow_right_hook: [window/showMessageRequest](#window_showMessageRequest) 
+* :arrow_left: [window/logMessage](#window_logMessage) 
+* :arrow_left: [telemetry/event](#telemetry_event) 
+
+Workspace
+
+* :arrow_right: [workspace/didChangeConfiguration](#workspace_didChangeConfiguration) 
+* :arrow_right: [workspace/didChangeWatchedFiles](#workspace_didChangeWatchedFiles) 
+* :leftwards_arrow_with_hook: [workspace/symbol](#workspace_symbol) 
+
+Document
+
+* :arrow_left: [textDocument/publishDiagnostics](#textDocument_publishDiagnostics) 
+* :arrow_right: [textDocument/didChange](#textDocument_didChange)  
+* :arrow_right: [textDocument/didClose](#textDocument_didClose) 
+* :arrow_right: [textDocument/didOpen](#textDocument_didOpen) 
+* :arrow_right: [textDocument/didSave](#textDocument_didSave) 
+* :leftwards_arrow_with_hook: [textDocument/completion](#textDocument_completion) 
+* :leftwards_arrow_with_hook: [completionItem/resolve](#completionItem_resolve) 
+* :leftwards_arrow_with_hook: [textDocument/hover](#textDocument_hover) 
+* :leftwards_arrow_with_hook: [textDocument/signatureHelp](#textDocument_signatureHelp) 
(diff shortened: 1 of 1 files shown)


protocol.md
```diff
@@ -1640,3 +1640,47 @@ interface RenameParams {
...
+
+* [window/showMessage](#window_showMessage) notification :arrow_left:
+* [window/showMessageRequest](#window_showMessageRequest) :arrow_right_hook:
+* [window/logMessage](#window_logMessage) :arrow_left:
+* [telemetry/event](#telemetry_event) :arrow_left:
+* [textDocument/publishDiagnostics](#textDocument_publishDiagnostics) :arrow_left:
+
+Notifications client &rarr; server
+
+* [workspace/didChangeConfiguration](#workspace_didChangeConfiguration) :arrow_right:
```
kaloyan-raev: This link does not work. I think you miss the anchor.

protocol.md
```diff
@@ -1640,3 +1640,47 @@ interface RenameParams {
...
+* [workspace/didChangeWatchedFiles](#workspace_didChangeWatchedFiles) :arrow_right:
+* [workspace/symbol](#workspace_symbol) :arrow_right:
```
kaloyan-raev: This is a request. It should be in the "Requests" section and the icon should be changed.

protocol.md
```diff
@@ -1640,3 +1640,47 @@ interface RenameParams {
...
+
+* [initialize](#initialize) :leftwards_arrow_with_hook:
+* [shutdown](#shutdown) :leftwards_arrow_with_hook:
+* [exit](#exit) notification :arrow_right:
+* [$/cancelRequest](#cancelRequest) :arrow_right:
+
+Notifications server &rarr; client
+
+* [window/showMessage](#window_showMessage) notification :arrow_left:
+* [window/showMessageRequest](#window_showMessageRequest) :arrow_right_hook:
```
kaloyan-raev: This is a request, not a notification.

vladdu:
Amended with the suggested changes. 

Changed the icons at the beginning, which gives a better feeling and grouped by area (workspace, window, textDocument). I feel it would be even better if there were different colors for the arrows, to be able to differentiate between c->s vs s->c. Maybe using real images instead of emojis is better, because the emojis are a github feature?

protocol.md
```diff
@@ -750,15 +750,15 @@ interface LogMessageParams {
...
+#### <a name="telemetry_event"></a>Telemetry Notification
 
 >**New:** The telemetry notification is sent from the server to the client to ask the client to log a telemetry event.
 
 _Notification_:
 * method: 'telemetry/event'
 * params: 'any'
 
-#### DidChangeConfiguration Notification
+#### <a name="workspace_didChangeNotification"></a>DidChangeConfiguration Notification
```
kaloyan-raev: Did you mean "workspace_didChangeConfiguration" here?

kaloyan-raev:
The new layout looks better.

I am personally fine with emojis. I read the protocol document directly on GitHub.

protocol.md
vladdu: Silly me... Fixed.

vladdu:
Hopefully the specification document will get generated from the schema, once #25 is implemented. The emoji names have no relation to their meaning.

felixfbecker:
I love this, but the ToC should be at the top, not the bottom

vladdu:
I put it at the end because it's a rather long list. The best solution would have been to have it as an aside, and with smaller text, but I think that would only be possible with raw HTML and that would make it a bit more difficult to maintain manually. 

Maybe at the top we could add a link to the TOC located at the bottom?

felixfbecker:
Well [GitHub supports `<details>`](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d), not sure about `<aside>`. But I still think having the ToC at the top would make it a lot more readable. I always have to use <kbd>Ctrl</kbd>+<kbd>F</kbd> to search for requests atm, and at the bottom the ToC is kind of useless.

vladdu:
If moving the TOC at the top, do you think it's best at the very top (before "Base Protocol") or before " Actual Protocol"?

felixfbecker:
The very top, after the introduction.

vladdu:
Ok, moved it.

dbaeumer:
Cool.

PR #119, opened 2016-11-09
Clarify what is meant by definition location for textDocument/definition

uforic:
Currently, the protocol states that:
```
The goto definition request is sent from the client to the server to resolve the definition location of a symbol at a given text document position.
```

It isn't clear, in the case of a struct Foo, whether this definition location should encompass _just_ the token `Foo`, or the entire body of the definition of the Foo struct:
```
type Foo struct {
    Bar bar
}
```

It seems like it should resolve to the entire body because:
- Visual Studio Code [hover preview depends on it](https://sourcegraph.com/github.com/Microsoft/vscode/-/blob/src/vs/editor/contrib/goToDeclaration/browser/goToDeclaration.ts#L329-334) spanning the body.
- It is more useful for many kinds of analysis e.g. determining function dependencies.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -1199,7 +1199,7 @@ interface ParameterInformation {
 
 #### Goto Definition Request
 
-The goto definition request is sent from the client to the server to resolve the definition location of a symbol at a given text document position.
+The goto definition request is sent from the client to the server to resolve the definition location of a symbol at a given text document position. The location is the entire range of the definition, e.g. the entire range of a function starting from its signature to the final closing code bracket.
 
 >**Changed:** In 2.0 the request uses `TextDocumentPositionParams` with proper `textDocument` and `position` properties. In 1.0 the uri of the referenced text document was inlined into the params object. 
 


dbaeumer:
@uforic VS Code hover doesn't depend on the definition request. I am not sure that we should spec this so strictly. Can you explain more why you think this is necessary.

uforic:
Ah, to be more specific - when you press cmd, or control in the case of Windows, while hovering in VSCode, you execute a definition request, and a preview of the definition appears , as in this [video](https://cl.ly/2f1O3C3I3i0y).

The code that renders that preview of the definition, which is [here](https://sourcegraph.com/github.com/Microsoft/vscode/-/blob/src/vs/editor/contrib/goToDeclaration/browser/goToDeclaration.ts#L329-334), is a conditional statement - there is a comment that says `// if we have a range, take that into consideration for the "to" position, otherwise fallback to MAX_SOURCE_PREVIEW_LINES.`

This seems to imply that in the best case scenario, the entire range of the definition would be returned in the `textDocument/definition` request, so that the preview could _just_ render the definition, and not naively render `MAX_SOURCE_PREVIEW_LINES` as it does in the fallback scenario.

But as for other reasons, here are some of analysis we've thought about that would be enabled by having the range encompass the entire definition:
* Detect if the code inside a method changes (i.e. someone decided to "watch" a definition, we'd like to know if a git diff touches lines inside that method).
* A visualization that shows what % of a file is taken up by which definitions
* If we issue a `textDocument/references` command, we'd like to be able to detect which definitions the references are occurring in. If we call `textDocument/documentSymbol` on the file where a reference occurs, if we had ranges, it'd be trivial to determine which definition the reference occurs in.

Thanks for the feedback - let me know what you think!

ljw1004:
There are three separate locations I might want to fetch:

1. The range of the symbol itself, i.e. just the "Foo" in "type Foo struct {...}"

2. The range of the declaration construct that defines Foo, i.e. the entire string "type Foo struct {...}"

3 .The range that includes doc-blocks, i.e. the entire string "/// Use this struct at your own risk\r\ntype Foo struct {...}"

Option [2] (which you had in your PR) doesn't seem that useful... if I'm doing a preview-on-hover then I'd like to see [3].

Option [3] has the problem that it's ill-defined for languages that have woolly conventions rather than crisp rules about where a docblock starts and ends. But it'd be the most useful for hover-preview.

Option [1] is the best-defined one of the three. As far as I can tell, it's what's used at the moment by language server implementations for textDocument/definition and also consistent with textDocument/documentSymbols and workspace/symbols.

If I'm right that option [1] is what implementations currently do, then we shouldn't *change* the protocol under their feet. If anything, either [2] or [3] should be *added* to the results.

uforic:
Re: Option 3 - Given that the response from textDocument/hover returns the marked string which includes doc-blocks, is it superfluous to also return those doc strings during textDocument/definition?

Definitely hear you re: the existing implementations point, most of the langservers I've seenreturn _just_ the token because highlighting that token after jumping to definition is the expected behavior. Seem like adding or modifying would be the way to go.

dbaeumer:
The major idea of the protocol is to get a decent UI integration not to have a correct language implementation. This is why in VS Code we use some fallback rules if the range is not provided or looks odd.

If we start to put a firm definition in and users will use the language server for program analysis (for which it is not designed right now) and then they will file issues servers not honoring the spec although the tools UI still does reasonable enough things with the result.

So in general I would like to keep the spec in the sense that the result is reasonable enough for a UI presented to a user which is able to decode the information in a meaningful way. I agree that this has to go in the spec in general. 

If we want to have a program analysis LSP then I thing we should have a different set of requests servers can opt in to provide.

Open for opinions.

object88:
I would suggest that `textDocument/definition` only include the token that represents the name (i.e., option 1).  While for a language like Go, an entire `struct` declaration may be relatively brief (a few lines, maybe up to a couple of dozen if there's a lot of members and comments), other languages are not so terse.  Consider C#, where the equivalent of a struct may span almost the entirety of a hundred- or  thousand-line file (i.e., [here](https://github.com/shadowsocks/shadowsocks-windows/blob/master/shadowsocks-csharp/Util/Util.cs)).  If a C# class definition is broken into several files (using partial classes), then the problem is conflated.

I can understand wanting to be able to get ranges where there are relevant implementations related to a token (i.e., where are its functions defined, etc), but I think that should be a separate, perhaps optional, request.

kdvolder:
Agree with @object88 statement. The basic info / range should be the actual token that has the name of the symbol. But... I can also see that some UI bits may work better if they also have the whole range of the definition (e.g. for a function like 'peek definition' it would be nice if the tool can only show the actual definition and not just some arbitrary amount of text around the name). 

So I think it would be useful to allow server implementation to include the 'whole range' as an optional attribute if it makes sense.

mickaelistria:
What about simply introducing a `caretPosition` into the Location type?

ljw1004:
@mickaelistria could you be more specific? What should LSP servers send as this caretPosition, and how should clients deal with it?

mickaelistria:
LSP servers should send the recommended position for the caret when returning this Location. For example, when returning a whole AST node, the caretPosition would for example return the position just before the name of the node in the declaration.
So clients would be using the range as they want (for example to show the node it in some hover, or to provide some operation such as "copy target node", "delete target node"...), but would strongly be recommended to place the caret where suggested when doing a "go to".

kdvolder:
I don't like the 'caret' suggestion. We are discussing about confusion about the meaning of the 'range'  ( @ljw1004 sumarized nicely what three reasonable interpretations are).

Introducing the extra 'caretPosition' does nothing to address that confusion. 

To make this more explicit, our language server has used Interpretation 1 and returns only the range of the name token of the symbol (i.e 'Foo' in Lucius example). It wouldn't help much to supply an extra caret position with that. And you shouldn't be proposing operations like 'delete definition' which would only delete the name.

ljw1004:
@mickaelistria I see. So let's spell out your proposal more precisely:

1. The three LSP requests that return a `Location` (GoToDefinition, FindReferences, DocumentSymbols) all do so because they're providing information about where a symbol is *used* or *defined*.

2. The `Location` type contains `Location.uri` and `Location.range`. Current LSP servers, when asked to give the `Location` of a *use-site*, invariably give the range of the use of the symbol identifier itself. Current LSP servers, when asked to give the `Location` of a *definition-site*, are mixed: most give the range just of the symbol identifier itself, but some give the range of a the whole declaration (which is a vague but practically useful concept).

3. You propose that the `Location` type contain a new optional element, `Location.identifierStartPosition`.

4. You propose that when Location is used to return a use-site, then `identifierStartPosition` will be the start of the symbol. You propose that when it's being used to return a declaration-site, then LSP servers will change so that `range` is always the full range of the declaration, and `identifierStartPosition` is the start position of the identifier.

5. You propose that for clients which wish to highlight the whole identifier, they should look at the start position, and then use their own custom language-specific regex (it's not part of LSP) to figure out how long that identifier is.


**EVALUATION**

It's not good to provide only the identifierStartPosition. It is important to provide the identifierRange. That's because there's no good robust way of determining the length of an identifier.

It's not good that `Location.range` should mean two such different things when used for a call-site vs a use-site. It would be better for `Location.range` to always refer universally to the range of the symbol in both cases. And then we can add a new optional field `Location.declarationRange`, used only for declaration-sites, which provides the full range of the declaration. (Again a vague concept and it's up to the language to decide what is the best range. I suggest that typically it shouldn't include preceding doc-blocks).

mickaelistria:
I understand that's what's being discussed is the meaning of `range` and I believe that the range should be the whole range of the AST node, and not only the identifier range. I simply don't get what value the users can get by having the range of the symbol selected in a declaration or a definition.
If a user does Ctrl+Click on a text in Eclipse IDE to move to declaration, just placing the caret where the symbol is defined as good (and even better) than selecting the name of the class for instance. Simply because when user does a Ctrl+Click on a text, the next action is extremely unlikely to require selection of the name of the symbol; users are more likely to scroll on the page, use Find References, some Rename or whatever. None of those make selecting th symbol name useful. Ie interpretation 1 described above is not more interesting than the other ones, and can be replaced by a `caretPosition` or similar without losing value in the protocol.
Since selecting symbol name seems useless to me, I believe the range is much more valuable being the whole node range as it enables more powerful feature to be implemented; and as a client of the LSP and user of IDEs, what I'm interested in "go to defintion" is only to place the caret at the right position.
IIRC, LSP4E deliberately ignores the range and only uses the 1st position to simply place the caret, because we found it better in practice.

So overall, I'm for range to be the whole node, and for an additional optional value to specify a caret position, because I believe it's what maps better user actions and it enables more powerful workflows.

object88:
I'm not sure how selecting an entire `struct` or `class` block would be more helpful when performing Find References or Rename, over selecting just the name token.  In the case of a language like C# of Java, where everything related to a class is defined within the class block, that seems to be very noisy.  Can you help me understand?

mickaelistria:
> I'm not sure how selecting an entire struct or class block would be more helpful when performing Find References or Rename, over selecting just the name token.

I'm asking the same question, but the other way round ;)

Range as entire node is more helpful IMO because it is more correct: the range of an element isn't the name of the symbol, it's the whole node.
I've seen the `documentSymbols` `range` attribute interpreted in 2 different ways: whole node and identifier only. The first one provides ability to build a Tree-based view of the object, to navigate in 2 directions between text and tree, while the 2nd one enables specific. Since then, I'm really convinced that having ranges thinking about nodes in general allows to get closer from the AST world which brings a lot of cool possibilities.
For example, in the case of Find References, let's assume you see the search results as a list. If the list contains the identifier only, then deleting an item of the list from the list (which would cascade to deletion of the identifier) drives to a syntax error; if the range represents the whole node, then deleting the search result from this list would delete the whole class/struct, which is a proper AST operation.
For `textDocument/definition`, I don't have a clear use-case in mind, except the one of "embedding" the definition in the hover to prevent user from switching to another editor to see the content (like Eclipse JDT allows). But this use-case is already pretty interesting. Anyway, my main argument would be that as ranges as whole node for documentSymbols are clearly more powerful and as I like consistency, I would be for ranges expressing the whole AST nodes whenever possible as I trust it would enable user stories plain identifier don't enable.
I think in general, dealing with ranges as close as possible from AST node enables more coarse-grain actions and alternative views on the documents.

>  that seems to be very noisy

That's more up to the tool to present it in a less noisy way. In Eclipse IDE, when a search result (Find Reference output is map as search results) is multiple lines, it simply shows the 1st few words, so it's not that noisier than showing only the work.

ljw1004:
@mickaelistria I agree that the range of the whole declaration has benefits. But I'm struggling to see how to generalize or formalize your idea. Here's a concrete example of ASTs:
```
Alpha.Beta.gamma(1);

[0]Call
+-[1]QualifiedIdentifier
| +-[2]QualifiedIdentifier
|   +-[3]identifier "Alpha"
|   +-[4]qualifier "Beta"
| +-[5]qualifier "gamma"
+-[6]argument "1"


class Beta {
  static Func<int> gamma
}

[7]ClassDecl
+-[8]identifier Beta
+-[9]member "static..."
```

In the first case, I did "Find all references" to find all references to symbol Beta. This might give the range for just node [4] in the AST, or it might walk up the tree to find an enclosing node [2] or [1] and give the range of that.

In the second case, I did "Go-to-definition" to find the definition of symbol Beta. This might give the range for just node [7], but you prefer it to walk up the tree to find the enclosing node [9] and give the range of that.


In general, do you propose that locations of *use-sites* should provide the range of the AST node for the identifier itself, but locations of *declaration-sites* should walk up the AST to find the smallest declaration node that contains the identifier?

kdvolder:
I don't think we should be arguing about whether it should be 'just the name' or 'the whole definition'. I'd rather argue that you probably want to allow the server to decide whether it wants to return one, the other, or both. And its up to the client to make the best use of whatever information, the server chosen to return.

If you only have the 'name range' then you can not really determine what range of text to show in the 'peek' window, so that's not great. But if you only have the 'whole range'  you can not easily determine where the name is, or perhaps, even *what* the name is. So I think that's not so great either.

Also I don't think we should be arguing to bake a 'single chosen meaning' into the spec. It was, so it seems, deliberately left vague. And so there are servers out there following different interpretations. You don't want to declare some percentage of those servers 'wrong' now, but rather give them a way forward with an augmented spec that tolerates their old use and allows them to become more precise about the intended meaning of what it is they are returning (and augment with additional info).

The caret position not having a good way to determine the 'end' of the name is a non-starter to me. You could just as well make that a 'name range' and if you really don't want to know where it ends, then you can choose to ignore that info. But if you only have the start, then there's no reliable way to determine the end.

dbaeumer:
Since the LSP is made to allow UI integration and not program analysis we should not try to spec this the same for GoToDefinition, FindReferences and DocumentSymbols. The result should allow for the best user experience and not for the program analysis correctness. I agree that we shouldn't spec a 'single chosen meaning'. We should rather spec what the tool usually does with the result. Something like this:

- GotoDefinition: the returned range is used to place the cursor at the starting point of the range when the user navigates to the definition. 
- DocumentSymbols: the returned range of the location is used to highlight the symbol in the document when the user browses through the list of symbols.
capture
- FindReferences: the return range is selected when the user picks a reference from the list.

If we need a different result for other user gestures then I think we should add additional request. The fact that VS Code uses goto definition for a peek preview is not correct and we should better have a different request (e.g. textDocument/peekPreview) to return a result that contains a meaningful range to be presented in a peek preview.

ljw1004:
@dbaeumer The DocumentSymbols API is currently used for [1] OutlineView (which would prefer to have entire range from start of declaration to end of `}`, so it can accurately reconstruct the tree).

Some clients might also use it just to [2] show a list of symbols and highlight whichever symbol the caret is currently inside, so it would prefer to have the whole range too.

It's only the use-case of a popup dialog [3] where you type in a name and scroll through the offerings (as you showed) which benefits from the range being only the symbol.

I think that to further the goal of LSP being a cross-editor API (rather than being narrowly tied to the UI of VSCode) it's important to address the [1/2]. They are after all different ways of displaying the list of available document symbols. I also think it's a non-starter to stick in a new LSP API for every single possible UI in every single editor. Here we have an API that's mostly perfect for [1/2] already, save for a small bit of clarification or an extra field.

dbaeumer:
@ljw1004 I only took VS Code as an example and the LSP shouldn't be tied to it. What I was trying to say is that GoToDefinition, FindReferences and DocumentSymbols use a Location underneath and that we can't specify what a Location is so we need to specify how the location is interpreted and therefore should look like in the above requests. 

But the tools should still meaningful behave if the result is not sticking to it. So I would rather add a outline request that has the concept of a clear hierarchy than enforcing that the result a document symbol request can be used to reconstruct a tree.

mickaelistria:
> So I would rather add a outline request that has the concept of a clear hierarchy than enforcing that the result a document symbol request can be used to reconstruct a tree.

I find this risky for LSP maintainability. If all unspecified behavior drive to a new operation, the number of operations will increase drastically, duplication of concerns is very likely to occur. It would IMHO be much more "proper" to work on the semantics of the existing data-types and operations, even if the only way forward is to deprecate some fields (such as `range`) in order to introduce disambiguized ones (`nodeRange`, `identifierRange`...).
I think LSP could do more with less operations if there was a path towards clarification or deprecation/replacement of ambiguous values. And doing more with less usually scales better in term of maintenance and adoption effort ;)

dbaeumer:
@mickaelistria I agree with this but it is not the right thing to do to spec a document symbol request in a way that clients can re-infer the hierarchy. If we need a hierarchical outline we should have a request for this.

The idea of the LSP is to talk UI abstractions not AST / compiler abstractions. If we start to add something like a node range and an identifier range we start to change this and IMO this make things more complicated considering that we want to be language agnostic. I still believe that abstraction over the UI across tools is easier than abstracting over AST / compiler concepts across programming languages.

I am all in for making the spec more clear. But I am still very confident that the clarification should happen in terms of UI abstractions not in terms of AST / compile abstractions.

PR #121, opened 2016-11-09
Add option for a preview line to textDocument/references

uforic:
I'm implementing an editor extension that will connect to a language server to display references to a symbol in Sublime.

As specified, the LSP textDocument/references call returns `Location[]`, which _just_ have uri and range information for the symbol references.

There are many situations where the user of a language server would want a preview of the reference. I think this use case might be common enough to warrant modifying LSP to allow for that.

Reasons to include preview with references:
- Efficiently accessing many files and extracting lines is harder from editor plugin environments than in a language server process.
- Many editors would need a preview for references - Sublime, VSCode, and many others already have this requirement. It seems common enough to warrant an API call.
- The [Microsoft TypeScript Sublime](https://github.com/Microsoft/TypeScript-Sublime-Plugin) plugin's bundled TypeScript language server [does this](https://github.com/Microsoft/TypeScript-Sublime-Plugin/blob/master/typescript/commands/references.py#L128).
- Servers implementing `textDocument/refences` need to open the file where the reference occurs already, so would they likely get this for free.

Reasons not to include preview:
- For editors like VSCode that support previews via the peek pane, the code and logic for acquiring chunks from different files already exists, so why duplicate code paths.

Open to suggestions, thanks!

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -231,10 +231,13 @@ interface Location {
 
 #### Preview Location
 
-Represents a location inside a resource, such as a line inside a text file.
+Represents a location inside a resource, such as a line inside a text file, and an optional preview.
 ```typescript
 interface LocationPreview extends Location {
-	contents: string;
+	/**
+	 * A preview of the text around the location.
+	 */
+	contents?: string;
 }
 ```
 
@@ -1240,14 +1243,10 @@ interface ReferenceContext {
 	 * Include the declaration of the current symbol.
 	 */
 	includeDeclaration: boolean;
-	/**
-	 * Include preview of line where reference occurs.
-	 */
-	includePreview: boolean;
 }
 ```
 _Response_:
-* result: [`Location`](#location)[] | [`LocationPreview`](#location-preview)[]
+* result: [`LocationPreview`](#location-preview)[]
 * error: code and message set in case an exception happens during the reference request.
 
 #### Document Highlights Request
--- a/protocol.md
+++ b/protocol.md
@@ -229,6 +229,15 @@ interface Location {
 }
 ```
 
+#### Preview Location
+
+Represents a location inside a resource, such as a line inside a text file.
+```typescript
+interface LocationPreview extends Location {
+	contents: string;
+}
+```
+
 #### Diagnostic
 
 Represents a diagnostic, such as a compiler error or warning. Diagnostic objects are only valid in the scope of a resource.
@@ -1231,10 +1240,14 @@ interface ReferenceContext {
 	 * Include the declaration of the current symbol.
 	 */
 	includeDeclaration: boolean;
+	/**
+	 * Include preview of line where reference occurs.
+	 */
+	includePreview: boolean;
 }
 ```
 _Response_:
-* result: [`Location`](#location)[]
+* result: [`Location`](#location)[] | [`LocationPreview`](#location-preview)[]
 * error: code and message set in case an exception happens during the reference request.
 
 #### Document Highlights Request


dbaeumer:
Good idea. But I think that the content field should be optional. I would force every language server to provide a content preview.

uforic:
Thanks for the feedback - I modified the pull request to return the LocationPreview, and made the contents field optional. Let me know if there are any other changes you'd like to see!

ljw1004:
This feels heavy-weight. What if it takes work to retrieve those previews (e.g. because the project is too large to fit in the LSP server's memory? or because they require additional disk lookup similar to autocomplete?)

I'm solving this problem by calling textDocument/hover. That's already the part of the LSP that's responsible for generating a preview for a given location in text. I don't think there should be a second place in LSP where previews are generated.

uforic:
Roger that - I'm wondering how the textDocument/hover idea works; the use case in mind is I find all textDocument/references to `struct Foo`, and there is a list of Locations[] that are returned.

Ultimately, the UX desires a preview of lines mentioned in the Location[]'s. I suspect if you called "textDocument/hover on the locations returned by textDocument/references, you'd get hover information for `struct Foo`, not a one-line preview of where the reference occurs.

It definitely takes work to retrieve those previews - hence it being optional. Justification for allowing this option is that however much work it is for the language server to request these files, it is less work and more reasonable for the LS to perform the work then the front-end client (which is usually more resource constrained).

rictic:
The LocationPreview.contents seems like it just duplicates the first line of the Range. Couldn't a client generate the contents if it wants them?

uforic:
Hey @rictic , definitely - in a client/server environment, in the case of 50 references found, do we want the client make 50 requests for entire documents and parsing out the right lines (which involves guessing the same encoding as the server), or would we prefer to have the server performing the work. Assumption here is that the client is more resource constrained, not the ideal place to be doing this kind of batch work, and it's common enough that we'd prefer the server to support it. Thoughts?

rictic:
Couldn't the client read documents on demand, as the user interacts? The protocol assumes that clients are able to access documents, decode them, and understand them in terms of lines and columns.

mickaelistria:
Agree with @rictic , the protocol already returns enough information for preview to be handled on client-side. For example, Eclipse IDE with LSP4E already gives preview of the line where reference is defined. I don't think this is something that needs to be added to the protocol.

dbaeumer:
I the server is really needs to be contacted for fetching the preview / parts of the document we might want to consider having a special request for this instead of having this embedded into find references. It might be useful in other places as well.

ljw1004:
For the scenario where files are stored remotely not locally, I think there are two completely separate functionalities:

1. Directory and file contents retrieval, and file-change notification, presumably provided by a server that's co-located with the files.

2. Language services based on files. This might be an LSP server which *uses* the file-server to fetch files, or it might be an LSP server which is *co-located* with the file-server and has access to the same files on disk, or it might be a *single JsonRPC server* which provides both LSP and file services.

I understand your motivating scenario, @uforic. For instance if we do find-references, and there are references from a 10kloc monster file like in Rosyn https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Generated/BoundNodes.xml.Generated.cs, then it'd be crazy to have to fetch 50 files that big merely to display previews.

But fetching file-contents is the job of the file server, not the LSP server. We shouldn't be adding arbitrary file-server-functionality into LSP.

Instead, to fulfill this scenario, I'd expect the fileserver to offer a service like "please fetch this range within this file". It would be fine to issue 50 requests to the server, one after the other, and thanks to HTTP2 it's no worse to send 50 requests then it is to send a single request that gets back an array of 50 strings.


__HOWEVER__ There's a problem that LSP has to knows the translation between byte-offset and line-numbers, but file-servers generally don't.

Maybe LSP should be sending byte-offsets in addition to line+columns, to support this kind of scenario.

Issue #18, opened 2016-06-28
Support semantic highlighting


hackwaly:
Like WebStorm and VS does: eg. Symbol is a type or parameter or namespace or unresolved ...

Textmate based grammars are hard to do this. Since we did support Diagnostics, why not support semantic highlighting?

dbaeumer:
Fully agree and we already discussed it since we would like to have it in VS Code as well.

daviwil:
I'd love to have this for the PowerShell extension also. We provide the ability to create "dynamic keywords" for the purpose of writing domain-specific languages.  Semantic highlighting would allow us to colorize those keywords in VS Code even though they aren't part of the PowerShell language spec. 

/cc @BrucePayette

smarr:
I was looking for that as well.

What's exactly the meaning of the current 'highlighting' for read/write/text (see [document highlights](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#document-highlights))?

I implemented that part but didn't see any for of visual feedback in the editor.

cdietrich:
DocumentHighlight is for "mark occurrences"

smarr:
@cdietrich sorry for hijacking the topic, but this is very unclear to me. How is document highlights used to realize a 'mark occurrences" feature? How does the read/write/text distinction fit in, and why does it only expect a single result in return? For mark occurrences, I'd expect to be able to return a collection, no?

Would be great if that could be clarified also in `protocol.md`.

cdietrich:
@smarr yes you are right. this makes no sense. i asume the return type should be an array.
if i have a look at vscode i can find
```
export interface DocumentHighlightProvider {
        provideDocumentHighlights(model: editor.IReadOnlyModel, position: Position, token: CancellationToken): DocumentHighlight[] | Thenable<DocumentHighlight[]>;
    }
```
thus looks like a bug in the protocol

smarr:
@cdietrich thanks, I'll open a separate issue.

svenefftinge:
Please find a proposal in PR #124

jpike88:
I’m guessing this is dead... anyone got any updates?

svenefftinge:
I am working on a new proposal for semantic coloring.

rcjsuen:
[...]

PR #124, opened 2016-11-15
Proposal for (Semantic) Coloring (see #18)

svenefftinge:
Here's a draft for the semantic coloring extension. It's a server side notification similar to publishDiagnostics, which leaves it up to the language server to decide when and if to update the coloring. 

We avoided using the term 'highlighting' because it is already used differently. The ids are an initial take, modelled after what we found in monarch with some additional modifiers.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -940,6 +940,39 @@ interface PublishDiagnosticsParams {
 }
 ```
 
+#### UpdateColoring Notification
+
+Coloring notifications are sent from the server to the client to update coloring information. This is mainly intended for semantic coloring and is usually triggered after a change event for an open document was changed and processed. It always contains all coloring information that should be put on top of the configuration based coloring on the client side.
+
+* params: `UpdateColoringParams` defined as follows:
+```typescript
+interface UpdateColoringParams {
+	/**
+	 * The URI for which coloring information is updated.
+	 */
+	uri: string;
+
+	/**
+	 * An array of ColoringInformation items.
+	 */
+	diagnostics: ColoringInformation[];
+}
+
+interface ColoringInformation {
+	/**
+	 * The range that should be highlighted on the client-side.
+	 */
+	range: Range;
+
+	/**
+	 * A list of highlighting style identifiers, that should be applied on
+	 * the range. Several styles could be merged on the client-side by 
+	 * applying all styles on the range. 
+	 */
+	ids: string[];
+}
+```
+
 #### Completion Request
 
 The Completion request is sent from the client to the server to compute completion items at a given cursor position. Completion items are presented in the [IntelliSense](https://code.visualstudio.com/docs/editor/editingevolved#_intellisense) user interface. If computing full completion items is expensive, servers can additionally provide a handler for the completion item resolve request ('completionItem/resolve'). This request is sent when a completion item is selected in the user interface. A typically use case is for example: the 'textDocument/completion' request doesn't fill in the `documentation` property for returned completion items since it is expensive to compute. When the item is selected in the user interface then a 'completionItem/resolve' request is sent with the selected completion item as a param. The returned completion item should have the documentation property filled in.
--- a/protocol.md
+++ b/protocol.md
@@ -944,6 +944,8 @@ interface PublishDiagnosticsParams {
 
 Coloring notifications are sent from the server to the client to update coloring information. This is mainly intended for semantic coloring and is usually triggered after a change event for an open document was changed and processed. It always contains all coloring information that should be put on top of the configuration based coloring on the client side.
 
+_Notification_
+* method: 'textDocument/updateColoring'
 * params: `UpdateColoringParams` defined as follows:
 ```typescript
 interface UpdateColoringParams {
--- a/protocol.md
+++ b/protocol.md
@@ -957,7 +957,7 @@ interface UpdateColoringParams {
 	/**
 	 * An array of ColoringInformation items.
 	 */
-	diagnostics: ColoringInformation[];
+	coloringInformation: ColoringInformation[];
 }
 
 interface ColoringInformation {
@@ -1002,8 +1002,8 @@ enum ColoringStyle {
 	Modifier_public = 22,
 	Modifier_private = 23,
 	Modifier_protected = 24,
-	Modifier_static = 22,
-	Modifier_final = 22
+	Modifier_static = 25,
+	Modifier_final = 26
 }
 ```
 
--- a/protocol.md
+++ b/protocol.md
@@ -971,7 +971,39 @@ interface ColoringInformation {
 	 * the range. Several styles could be merged on the client-side by 
 	 * applying all styles on the range. 
 	 */
-	ids: string[];
+	ids: int[];
+}
+
+enum ColoringStyle {
+	
+	Identifier     = 1,
+	Entity         = 2,
+	Constructor    = 3,
+	Operators      = 4,
+	Tag            = 5,
+	Namespace      = 6,
+	Keyword        = 7,
+	Info_token     = 8,
+	Type           = 9,
+	String         = 10,
+	Warn_token     = 11,
+	Predefined     = 12,
+	String_escape  = 13,
+	Error_token    = 14,
+	Invalid        = 15,
+	Comment        = 16,
+	Debug_token    = 17,
+	Comment_doc    = 18,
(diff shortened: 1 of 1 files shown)


smarr:
This would be a great feature to have!

I would expect the ids to be the part that requires most discussion.
Should the ids be explained? This part is also highly language specific, so, I suppose it should cover a combined set of elements for various languages and still be somewhat manageable.

One thing I would like to be able to do is to distinguish field accesses from method invocations based on data from the server side, because in my language it is a unified syntax and coloring could provide a real benefit to code readability.

Not sure why you chose to model different modifiers separately. Is this intentional?

svenefftinge:
> One thing I would like to be able to do is to distinguish field accesses from method invocations based on data from the server side, because in my language it is a unified syntax and coloring could provide a real benefit to code readability.

Do you mean something like 'read', 'write', 'declaration'? Would make sense indeed.

> Not sure why you chose to model different modifiers separately. Is this intentional?

The idea was to allow combinations of orthogonal things, like "method & static & protected", or "field & private".

Or do you mean the 'modifier_' prefix? That is just because I thought it would be better explain the intended meaning, not sure if that really helps, though.

smarr:
> Or do you mean the 'modifier_' prefix? That is just because I thought it would be better explain the intended meaning, not sure if that really helps, though.

I was wondering why you have `Modifier_*` instead of just `modifier`. The ones in your list are also very language specific. So, perhaps it would be better to have `modifier` and then in addition `visibility` (for public, ...), `storage` (for static), `changeability` (final, readonly, ...)?

Perhaps the notion of ids is also more a notion of "tags"? I assume here that one can combine tags, while ids are not usually combinable. But that's more about terminology.

dbaeumer:
@svenefftinge thanks for the PR. We discussed semantic coloring in the VS Code team as well and we came to the conclusion that, to make it work nicely, it needs to support some delta encoding. So when a files open all sematic colors are computed. When the file is change only delta are computed and sent back to the client. Have you thought about how this could work ?

freebroccolo:
I think it would be great to have some sort of more official support for compiler-based colorization in the language server protocol.

The following is specific to vscode, but I have a prototype implementation of compiler-based highlighting for the agda language at [vscode-agda](https://github.com/freebroccolo/vscode-agda). It uses custom lsp requests and decorations on the client side. Agda allows very flexible syntax (operators can essentially modify the parsing rules) so compiler-based colorization is the only way to provide accurate highlighting. Here is an example of what it can produce (all highlighting is computed by the server):

agda-highlighting

One issue I have run into several times is the need to be able to recover from the client the live decoration positions in a dirty buffer. Sometimes this is needed to work around limitations with the decoration approach (e.g., highlighting must be saved/refreshed on visibility changes) but sometimes this is needed in order to compute new colorization data depending on certain actions in a particular editing context. This is even more important if the server does not have good support for incremental document synchronization.

For vscode, there is currently there is no way to do this through the API so one would have to maintain their own position deltas for the decorations as edits are made. This has negative implications for performance and maintainability.

Being able to send colorization deltas as @dbaeumer suggests would be ideal but I think that approach should be optional, similar to incremental vs full sync for edits. In my prototype, for instance, I am not using a delta encoding because `agda` does not provide good support for that and it would be much more work and ultimately slower to try and compute deltas on the server.

As for the `ColoringStyle` ids, I am not sure how useful it is to include them in the protocol.  Only a few of the proposed ones would even fit my use case. [Here](https://github.com/freebroccolo/vscode-agda/blob/master/src/client/window.ts#L11-L37) is a list of the ones I use which are determined by the compiler.

svenefftinge:
Yes, incremental update seem to be a needed optimization for large files, but it for sure complicates things which is why I would prefer that to be optional. 

How about an optional property `invalidationRanges : Range[]`, that tells the editor which coloring positions should be removed before applying the new ones? If it is undefined we treat it as a full replacement.

We will need to assume that the server and the client move positions according to changes.
Otherwise we would always need to update all ranges after a certain edit, reducing the possible optimization significantly.

paulotorrens:
@svenefftinge, I believe that the protocol should be able do handle more exotic languages. The ids in your proposal seem perfectly fit for an object oriented language, but may not have enough options for some languages. E.g., heredoc comments in Perl or Ruby (the later having ["end of file data"](https://shifteleven.com/articles/2009/02/09/useless-ruby-tricks-data-and-__end__/), for which a language server writer might want to give special attention).

One cool feature, for example, would be level highlighting, as in:

</img>

Ideally, instead of ids, maybe it should accept strings instead of predefined numbers, which could be handled by the client from a map of known names (e.g., "keyword"), or even be handled by the client as CSS classes (as I'm working on a prototype of mine).

What do you think?

MarkZ3:
Any news about this? I'm interested in implementing this in a server soon :)

Lisoph:
### TextMate?

Instead of a `ColoringStyle` enum, why not simply send **TextMate identifiers** as strings? The client side could very easily use these to drive it's colorization, assuming it's based on a TextMate compatible highlighter.

Basically, allow the server to act like a TextMate "tokenizer". Given source files as an input, it computes a nested list of identifiers for each file. These identifiers can then be easily matched against .tmTheme files.

This would allow for a very reasonable compability with most languages and color themes.

#### Little excursion to TextMate grammars
Let's say I'm writing a server for the acient language *Baan*. The server could transmit identifiers such as `keyword.4gl.main.table.section.baan`. That identifier is very specific to the language Baan, but because of the nature of TextMate grammars, 99% of all existing .tmThemes would still work properly since they would match against the `keyword…` part of the identifier.

Of course to save performance, only the identifiers of regions affected by a source code change could be transmitted to the client. Further optimizations are possible too.

What do you guys think?

Lisoph:
Does anyone have any comments on my suggestion?

Xanewok:
Just to chime in about possible usage, it seems that indicating conditional compilation would benefit greatly from this (e.g. C/C++ `#if 0`/`#ifdef`, Rust `#[cfg(...)]`, tracked over at  https://github.com/rust-lang-nursery/rls/issues/384).

Lisoph:
@Xanewok indeed it would! Making code between `#if WEATHER_IS_NICE_TODAY = 1` and `#endif` appear grayed out, would be very useful. Same goes for Rust and languages with similar constructs.

This would also work well together with my TextMate proposal.

Lisoph:
Addition to my TextMate proposal: The [TextMate naming conventions page](http://manual.macromates.com/en/language_grammars#naming_conventions) lists all the de facto standard "identifiers" used by most .tmTheme and .tmLanguage files.

svenefftinge:
> Just to chime in about possible usage, it seems that indicating conditional compilation would  benefit greatly from this (e.g. C/C++ #if 0/#ifdef, Rust #[cfg(...)], tracked over at rust-lang-nursery/rls#384).

Couldn't that be entirely handled by the language server?

+1 for reusing the textmate identifiers.

Xanewok:
> Couldn't that be entirely handled by the language server?

It could and it should but from what I understand being able to display differently, e.g. grey out, code that won't be compiled needs the additional proposed notification from the server to the client, no?

svenefftinge:
Yes, sounds right. I think the feature is a bit too specific to a certain kind of language to be added to the protocol, though.

deathaxe:
Why not use `scoping` instead of `coloring`? This is what Sublime Text uses. It creates scopes for source tokens, which are used by color schemes in the second place to assign a color to.

jpike88:
I’m guessing this is dead... anyone got any updates?

Lisoph:
I'm afraid this isn't going anywhere since this PR doesn't follow the [contribution guide](https://github.com/Microsoft/language-server-protocol/blob/master/contributing.md).

We need to define an protocol extension as described and also provide a reference implementation. I'd love to do this myself, but I no good with TypeScript nor node.

MrYurihi:
I thing it would be more natural to send a string array of scopes instead of an int array seeing as many text editors use a scope format (Atom, VSCode, TextMate, etc)

svenefftinge:
> I thing it would be more natural to send a string array of scopes instead of an int array seeing as many text editors use a scope format (Atom, VSCode, TextMate, etc)

Yes, agreed. We are working on a new proposal, that will use Textmate scopes.

svenefftinge:
Closing in favor of https://github.com/Microsoft/vscode-languageserver-node/pull/367

PR #131, opened 2016-11-23
Add Annotation in the CompletionItemKind enum

kaloyan-raev:
Annotations are popular kind of completion items in Java, Scala and other programming languages.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -1101,7 +1101,8 @@ enum CompletionItemKind {
 	Snippet = 15,
 	Color = 16,
 	File = 17,
-	Reference = 18
+	Reference = 18,
+	Annotation = 19
 }
 ```
 * error: code and message set in case an exception happens during the completion request.


dbaeumer:
I like adding this but this needs adoption on clients sides as well.

ljw1004:
Shouldn't it also be added in the SymbolKind enumeration? e.g. if you do a project-wide symbol search and the result is an annotation.

(I think it'd be good for users if they have consistent icons between autocomplete and project-wide-symbol-search).

dbaeumer:
Closing in favour of https://github.com/Microsoft/language-server-protocol/issues/344 which contains Annotation.

KimBurgess:
It looks like this may have been closed off in favour of a (still open, long running) discussion, rather than alternate PR. Is there a current recommended / well used alternative `SymbolKind` that can be used for annotations?

PR #166, opened 2017-01-23
Added ComplexDiagnostic

KalitaAlexey:
Hi everybody.

Currently there is no way to show complex diagnostics.

Rust have very useful and detailed messages.

For code:

```rust
fn main() {
    let mut x = &5;
    {
        let y = 5;
        x = &y;
    }
}
```

It gives very good explanation:
```
error: `y` does not live long enough
 --> main.rs:6:5
  |
5 |         x = &y;
  |              - borrow occurs here
6 |     }
  |     ^ `y` dropped here while still borrowed
7 | }
  | - borrowed value needs to live until here
```

I am using VSCode.
Rust Language Server (RLS) is currently being developed.
RLS can send all of aforementioned messages, but there will be problems when multiple messages have same range.

rls

There are errors on both 7 and 8 lines.
They point to both 9 and 10 lines.

This is how it works with code lenses:

screenshot 1

I'd like to have a way to show only primary message and if a user hover on the message to show all secondary messages.
Currently there is no way to show such kind of messages in VSCode.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -375,6 +375,24 @@ namespace DiagnosticSeverity {
 }
 ```
 
+#### ComplexDiagnostic
+
+Represents a complex diagnostic, such as a compiler complex error or warning. ComplexDiagnostic objects are only valid in the scope of a resource.
+
+```typescript
+interface ComplexDiagnostic {
+	/**
+	 * A valuable diagnostic
+	 */
+	mainDiagnostic: Diagnostic;
+
+	/**
+	 * An array of helpful, but less valuable diagnostics
+	 */
+	slaveDiagnostics: Diagnostic[];
+}
+```
+
 #### Command
 
 Represents a reference to a command. Provides a title which will be used to represent a command in the UI. Commands are identitifed using a string identifier and the protocol currently doesn't specify a set of well known commands. So executing a command requires some tool extension code.
@@ -1678,6 +1696,11 @@ interface PublishDiagnosticsParams {
 	 * An array of diagnostic information items.
 	 */
 	diagnostics: Diagnostic[];
+
+	/**
+	 * An array of complex diagnostic information items.
+	 */
+	complexDiagnostics: Diagnostic[];
 }
 ```
 


KalitaAlexey:
@jonathandturner,
I have created an issue allowing RLS to show beautiful messages.
What do you think about it?

wiz:
`slaveDiagnostic`? In 2017?..

Anyway, why the new type and not optional field in the original message?

KalitaAlexey:
@wiz,
What's wrong with `slaveDiagnostic`?
The new type because there is no such type.
Should I add `?` to make a field optional?

KalitaAlexey:
@wiz,
Do you link `slaveDiagnostic` with slavery? Is **secondary** better?

wiz:
> The new type because there is no such type.

https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#diagnostic

There's message already, but now notion of how editor should behave with respect to multiline string.

If you've tried sending some `\n`s there and it doesn't work, a new `details?` field could be added.

KalitaAlexey:
@wiz,
Look at the provided information.
Diagnostics are on different lines.

wiz:
> Currently there is no way to show such kind of messages in VSCode.

What does it show instead?

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

KalitaAlexey:
@wiz 
>> Currently there is no way to show such kind of messages in VSCode.

>What does it show instead?

It show each message separately. I described when it is a problem:
*...there will be problems when multiple messages have same range.*

```rust
fn main() {
    let mut x1 = &5;
    let mut x2 = &5;
    {
        let y1 = 5;
        x1 = &y1;

        let y2 = 5;
        x2 = &y2; 
    }
}
```
Output:
```
error: `y1` does not live long enough
  --> main.rs:10:5
   |
6  |         x1 = &y1;
   |               -- borrow occurs here
...
10 |     }
   |     ^ `y1` dropped here while still borrowed
11 | }
   | - borrowed value needs to live until here

error: `y2` does not live long enough
  --> main.rs:10:5
   |
9  |         x2 = &y2;
   |               -- borrow occurs here
10 |     }
   |     ^ `y2` dropped here while still borrowed
11 | }
   | - borrowed value needs to live until here
```

wiz:
> It show each message separately

You'd better start posting screenshots :smile: I understand RLS messages, but fail to imagine the result in the editor itself...

KalitaAlexey:
@wiz,
Okay. I will attach screenshots.

KalitaAlexey:
I have added a screenshot showing the problem.

KalitaAlexey:
@nrc,
Hi.
I am inviting you here to hear your opinion.

wiz:
> I have added a screenshot showing the problem.

Still can't see what message does the editor currently show on hover or anywhere else...

KalitaAlexey:
@wiz,
I'll add more screenshots.
I didn't think that it is needed.

dbaeumer:
I see the need for more complex diagnostics given the analysis modern languages perform. However using commands for this is IMO the wrong solution. VS Code for example has already support for showing inline error message (as other editors have).

capture

In addition something like a complex diagnostic message must be handled by a problem view as well. This means that publishing diagnostic should support such complex diagnostics and the editor / client should handle them in an appropriate and consistent way in its user interface.

mickaelistria:
Rather then introducing a new type, couldn't a diagnostic have an optional reference to a `master` or `cause`other diagnostic?

KalitaAlexey:
@mickaelistria,
Why?

nrc:
From our point of view (implementing the RLS) we would rather experiment by adding custom fields to the `Diagnostic` interface to begin with. I don't think we need to make official changes to the protocol to  make this work (aiui, VSCode and other editors should ignore unexpected fields, not error out). Once we've found what works, we'll report back. I think making any changes at this stage would be premature.

For reference, our initial plan is adding `label: string` and `children: ChildMessage` where

```
interface ChildMessage {
    range: Range,
    label: string,
}
```

I expect that the new `label` field is a bit Rust specific, so an alternative plan is to add:

```
interface LabelledRange {
    range: Range,
    label: string
}
```

Then change `range: Range` in Diagnostic to `range: Range | LabelledRange` and add `childRanges: [Range| LabelledRange]`, or something like that. As I say, we're experimenting :-)

jonathandturner:
Just to follow-up - since we want to keep the structures compatible, we could do something like this:

```
interface LabelledRange  extends Range {
  label: string
}
```

This way, structurally we stay compatible with the shipping version of the error message object, but we can add a few additional fields that would be seamlessly ignored by LSP compatible plugins.

Actually, because we're still so early in the thinking here, it makes sense to close this issue until we've had time to create and test out a real proposal.

KalitaAlexey:
@jonathandturner,
Okay. I will close it for now.

dbaeumer:
I like the proposal very much to get something working and then to spec the LSP. If you do the prototyping in VS Code please consider the problems view and the F8 feature (Goto next Error or Warning) as well.

KalitaAlexey:
I am not going to open this PR since probably something will change.
Instead of it, I would propose to allow the client to notify about hovering:

* text/willShowHover with the range
* text/didHideHover with the range

This way our language server would be able to show code lenses on starting hovering and to remove code lenses on stopping hovering.

PR #170, opened 2017-01-30
Introduce DocumentUri concept.

bolinfest:
In the protocol, interface fields that are meant to specify the URI of a
document are simply declared as `string`. This underspecifies the requirements
of these fields because the contents of the string must be able to parse as a
URI (at a minimum, using the `vscode-uri` npm module).

This introduces a type alias, `DocumentUri`, that tightens these guarantees
without changing the wire protocol. It updates what I believe are all of the
relevant interfaces to specify `DocumentUri` instead of `string`, as
appropriate.

Note that this type is deliberately named `DocumentUri` rather than simply
`Uri` to convey that it is a `URI` that identifies a `TextDocument` in the
LSP object model (as opposed to an arbitrary URI).

I am by no means tied to the name of the type alias: I just want this concept to
be formalized in the specification so it is possible to create a proxy that
can reliably rewrite all `DocumentUri` parameters such that a local editor can
make requests to a remote language server even though different URIs may be used
on both sides of the connection to refer to the same document.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -1,9 +1,9 @@
-# Language Server Protocol 
+# Language Server Protocol
 
 This document describes version 3.0 of the language server protocol. Major goals of the 3.0 version are:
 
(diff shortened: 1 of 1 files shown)


dbaeumer:
@bolinfest Thanks for making this easier to understand.

Issue #141, opened 2016-12-08
Check partial answers for ongoing requests


vladdu:
For long-running requests, it makes sense to show partial results in the UI (for example when searching for all references in a workspace, but even for completion). Streaming them from the server is not a very good idea, but the client could send progress requests whenever it wants and receive the current results, updating the UI. Kind of like `cancelRequest` but without cancelling.

felixfbecker:
I have made a proposal for progress in #70

vladdu:
This is more than progress report. I'd like to be able to get a list of partial results for an ongoing request. This should be something the client requests.

kdvolder:
> Streaming them from the server is not a very good idea,

Why not? Personally I very much like the idea of having a fully async and incremental mechanic for sending completions (for example). 

Having this allows for deliver any results as soon it has them rather having to make an awkward decision around how much time is 'reasonable' to try and compute as many results as possible before returning an incomplete list. This often means you end up returning 0 completions even if the user was willing to wait a bit (or just stopped typing for a moment to reflect).

vladdu:
I think also that it would be neat with a streaming API, fully async. It has the downside is that it offers no flow control. Also, I believe that clients might have trouble handling it properly, thus making it less likely to get implemented. I started with a less ambitious suggestion, hoping to get it accepted easier.

I have absolutely no objection against a fully async protocol, when it is supported by both the client and the server. There will have to be a fallback solution for participants that need flow control or can't handle the async streams of events.

dbaeumer:
With the new capability flag support in 3.0 I can imagine a flag 'supportsResultStreaming' on the completion capability to indicate that the client accept streamed results. Then the server can stream results.

felixfbecker:
> With the new capability flag support in 3.0 I can imagine a flag 'supportsResultStreaming' on the completion capability to indicate that the client accept streamed results. Then the server can stream results.

How exactly would that work? Streaming depends entirely on the client parsing the response as a JSON stream, for example if it is an array emitting each array item instantly and not waiting for the whole response to parse it completely.

vladdu:
Not necessarily, I think. The server could also send multiple answers to a request, with a flag saying if it's done or not. There should be a merging policy for the non-array values. This way, it works fine even if a response would contain two arrays that are computed in parallel (there isn't such a response now, but there could be).

dbaeumer:
[...]

PR #182, opened 2017-02-14
Add streaming support

felixfbecker:
This is a proposal for adding "streaming" support to LSP, more specifically: The use case of wanting to return parts of a result before the full result is finished (#141). This is especially desirable for methods that can return lots of results like `workspace/symbol`, `textDocument/completion` and are therefor often expensive, even when the UI can only shows a subset of the full response anyway.

There are a couple of approaches to this:
- **Using a parser that can parse JSON as a stream**
  This is not really viable because
    - the server can only "modify" one "object" in the result at a time, not update information at any path as information is available
    - the protocol requires to send a `Content-Length` header, so the sender needs to know the full content in advance
- **Polling the server for a progress / updates**
  Polling results in unnecessary overhead. With the request the client expressed clear interest in the result, so the server should provide what it has as soon as possible.

Introducing a new method (`$/partialResult`) that the _server_ calls on the _client_ is therefor the best approach. This new method should be passed a request ID the server is returning partial results for and the partial data itself.

The data structure used to represent the partial result should work for all methods. In the simplest case, that means adding items to a result array. That should also work if the array is nested inside an object - for example, `textDocument/completion` returns an object with `isIncomplete` and `items`. The language server should be able to fill out the `isIncomplete` as well as add new items to the `items` array with partial results. The language server should also have the ability to provide partial objects, then fill out missing properties later. For example, it should be able to provide `Location`s with just URIs initially or `SymbolInformation`s with just the name, then fill out the `range` later because it might be expensive to compute, not desirable to keep in memory and is not needed to show the user initially. The client could then already make use of the partial information.

Just trying to "merge" multiple objects/arrays together will not suffice for this as there would be no way to address specific array indices or to append to an array.

In addition, using a well-defined standard would have the benefit of having libraries available in many languages that ease implementation.

This is solved by [JSON Patch](http://jsonpatch.com/). A JSON Patch is nothing but an array of _operations_ which modify a value at a certain path. In particular, it is possible to add more items to arrays as well as add properties to objects no matter how nested the value is inside the result.

The client would just continuously apply the operations to a result and then for example rerender the UI. The UI might have to be modified so it handles undefined values, for example by showing _Loading..._ where it would normally show certain properties or filtering items that it cannot display because of incompleteness. It might also cancel the request after it has received the information it was looking for.

The final response indicates the result is complete. If the receiver opted into streaming through the `Server`/`ClientCapability` and the server sent at least one partial result notification then the response result can just be `null` and is ignored by the client. Applying all patches sent through `$/partialResult` should yield to the exact same result as a not streamed response.

JSON Patch has libraries for almost every language including JS, Python, PHP, Ruby, Perl, C, Java, C#, Go, Haskell and Erlang. This should make implementation easy for clients and servers.

Simple example:

```ts
// --> Client sends a references request
{ id: 1, method: "textDocument/references", params: {} }
// Result at the client: undefined

// <-- Server sends a partial result that initializes the result to an empty array
{
	method: "$/partialResult",
	params: {
		id: 1,
		patch: [{ op: "replace", path: "", value: [] }]
	}
}
// Result at the client: []

// <-- Server sends zero to n partial results that append a Location to the array
{
	method: "$/partialResult",
	params: {
		id: 1,
		patch: [{ op: "add", path: "/-", value: { uri: "file:///a/file", range: { start: { line: 1, character: 2 }, end: { line: 1, character: 3 }}}]
	}
}
// Result at the client:  [{ uri: "file:///a/file", range: { start: { line: 1, character: 2 }, end: { line: 1, character: 3 }}}]

// <-- Server sends the final response
{
	id: 1,
	result: [{ uri: "file:///a/file", range: { start: { line: 1, character: 2 }, end: { line: 1, character: 3 }}}]
}
```

We experimented with this proposal at Sourcegraph in two language servers and are working on a VS Code client implementation.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -23,6 +23,7 @@ General
 * :leftwards_arrow_with_hook: [shutdown](#shutdown)
 * :arrow_right: [exit](#exit)
 * :arrow_right: [$/cancelRequest](#cancelRequest)
+* :arrow_left: [$/partialResult](#partialResult)
 
 Window
 
@@ -237,6 +238,37 @@ interface CancelParams {
 ```
 
 A request that got canceled still needs to return from the server and send a response back. It can not be left open / hanging. This is in line with the JSON RPC protocol that requires that every request sends a response back. In addition it allows for returning partial results on cancel.
+
+#### <a name="partialResult"></a> Streaming Support
+
+When a language server or client is able to provide partial information before having the final result,
+it can send `$/partialResult` notifications to return parts of the result.
+
+_Notification_:
+* method: '$/partialResult'
+* params: `PartialResultParams` defined as follows:
+```typescript
+interface PartialResultParams {
+	/**
+	 * The request id to provide parts of the result for
+	 */
+	id: number | string;
+
+	/**
+	 * A JSON Patch that represents updates to the partial result as specified in RFC6902
+	 * https://tools.ietf.org/html/rfc6902
+	 */
+	path: JSONPatch;
+}
+```
+
+The receiver can listen for partial result notifications with the request ID and use the provided JSON patches to build up a result.
+The JSON Patch can contain every possible
+It is free to do whatever it wants with the partial result.
+The object is indicated as complete by the final response to the request, which should still contain the final result for BC.
+The result is _not_ required to conform to the result interface while still being built.
+For example, a language server might first return an array of several `Location`s with just the URI set and then fill out the ranges with patches.
+Applying every patch should result in the same result as send in the final response.
 
 ## Language Server Protocol
 
--- a/protocol.md
+++ b/protocol.md
@@ -265,7 +265,7 @@ interface PartialResultParams {
 
 The receiver can listen for partial result notifications with the request ID and use the provided JSON patches to build up a result.
 The result is _not_ required to conform to the result interface while still being built.
-For example, a language server might first return an array of several `Location`s with just the URI set and then fill out the ranges with patches.
+For example, a language server might first return an array of several `SymbolInformation`s with just the name and kind set and then fill out the locations with patches.
 The result is indicated as complete by the final response to the request, which, assuming the client expressed support for streaming through `ClientCapabilities`, does not contain any data itself (`result` must be `null`).
 The same applies for requests sent from the server to the client (see `ServerCapabilities`).
 A result built by applying all patches sent through `$/partialResult` should eventually yield in the exact same result as a non-streamed response result.
--- a/protocol.md
+++ b/protocol.md
@@ -266,7 +266,7 @@ interface PartialResultParams {
 The receiver can listen for partial result notifications with the request ID and use the provided JSON patches to build up a result.
 The result is _not_ required to conform to the result interface while still being built.
 For example, a language server might first return an array of several `Location`s with just the URI set and then fill out the ranges with patches.
-The result is indicated as complete by the final response to the request, which can contain just `null` as value for `result` if the client expressed support for streaming through `ClientCapabilities`.
+The result is indicated as complete by the final response to the request, which, assuming the client expressed support for streaming through `ClientCapabilities`, does not contain any data itself (`result` must be `null`).
 The same applies for requests sent from the server to the client (see `ServerCapabilities`).
 A result built by applying all patches sent through `$/partialResult` should eventually yield in the exact same result as a non-streamed response result.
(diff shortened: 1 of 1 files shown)


nicksnyder:
See https://github.com/Microsoft/vscode/issues/20010 for an umbrella issue that I have created to track the associated implementation details.

Specifically these two PRs demonstrate an implementation of this proposal in VS Code
- https://github.com/Microsoft/vscode/pull/20346
- https://github.com/Microsoft/vscode-languageserver-node/pull/164

felixfbecker:
ping @dbaeumer

ljw1004:
Conceptually, diagnostics are a stream. (The protocol currently lacks a way for the server to indicate "there are no further diagnostics left to report" but it should have one).

Can you think of any way to unify diagnostics with streamed results?

felixfbecker:
@ljw1004 could you explain the where you see the streaming in diagnostics? Diagnostics are reported per-file and clear previously reported diagnostics for that file. They are typically only reported when you open a file, not for all files, so there is usually no real known point at which all diagnostics for all files are reported.

ljw1004:
> They are typically only reported when you open a file, not for all files

Why do you say that? In C# and Visual Studio, when you edit a file (e.g. changing the signature of a method), then it works through the *entire* project to find every single place that might be affected. It reports errors including for files that haven't yet been opened.

When you work on the Roslyn codebase itself in Visual Studio (it's a fairly large C# project) then it sometimes takes up to a minute to figure out all the errors in other files. In more moderately-sized projects I commonly saw 10+ seconds to figure out all the errors in other files.

For our own team, we also display errors throughout the entire project, but our project is even bigger so it takes a bit longer.

dbaeumer:
Regarding the ping https://github.com/Microsoft/language-server-protocol/pull/182#issuecomment-285037398: given the priorities we currently have for VS Code I can't commit on giving reasonable feedback on this pull request in the near future. I agree that streaming support is a cool feature for the LSP but it requires quite some work to get this right especially for clients. May be we should start with a progress specification first which could mitigate some of the streaming use cases.

Regarding markers: language servers can provide markers for all resources in a workspace. There is no restriction to only handle open resources. This is why diagnostics are actually 'stream based' on a resource level and are notification send from the server to the client and not request results.

matklad:
Note that Data Analysis server support streaming in a pretty general way, without using JSON patch, streaming parsers or other complicated tools. An example would be symbol search: https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#request_search.findTopLevelDeclarations

The way it works is that, for "streaming" requests, the server answers immediately with a `SearchID`, and then proceeds with sending notifications with this specific search id.

Each notification usually contains the *full* set of items (so, newer ones overwrite older ones), but in a select cases (workspaceSymbol being one of these) the results from all notifications are concatenated.

dbaeumer:
Please see also #786

dbaeumer:
Closing the issue. Corresponding support has been speced in dbaeumer/3.15 branch.

ljw1004:
@dbaeumer I'm looking in the [dbaeumer/3.15 branch](https://github.com/microsoft/language-server-protocol/compare/dbaeumer/3.15...master?expand=1) but I see nothing there about streaming - the only mention of "stream" is for the LSIF exporter, and there seems to be no mention of JSON-Patch or progress or partialResult. Am I just looking in the wrong place?

KamasamaK:
@ljw1004 Look at the [specification.md](https://github.com/microsoft/language-server-protocol/blob/dbaeumer/3.15/specification.md). It is achieved by supporting partial results.

ljw1004:
Ah, thanks @KamasamaK - I'd simply been looking in the wrong place.

dbaeumer:
@ljw1004 I published the upcoming 3.15 spec on the Web/UI here: https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/

The progress relevant sections are:

https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#progress

https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#partialResults

yyoncho:
Are there any language servers implementing this feature?

yyoncho:
> Are there any language servers implementing this feature?

Someone? Or at least is there a language server planning to implement it anytime soon?

PR #198, opened 2017-03-08
DidChange might only have new text

ljw1004:
Currently the spec says that `TextDocumentContentChangeEvent.text` contains the new text of the *document*.

I assume that, if the server requested incremental changes and if the client obliged by sending a range+rangelength, then the `text` field merely contains the new text of the *range*.


dbaeumer:
Correct !

PR #199, opened 2017-03-09
Change "utf8" to "utf-8" with a hyphen

ljw1004:
The charset name is "utf-8" with a hyphen. I think that "utf8" without a hyphen doesn't even exist in the wild.

https://tools.ietf.org/html/rfc3629 - section 8
https://tools.ietf.org/html/rfc5987 - section 3.2.1


dbaeumer:
Agree. It is the constant usually used in node.

PR #204, opened 2017-03-22
"experimential" -> "experimental"

arxanas:
Fix typo. None of the actual spec text appears to say `experimential` instead of `experimental`, so the changes should be purely that of documentation.


dbaeumer:
@arxanas thanks for the PR. The typo got already fixed through another PR.

Issue #83, opened 2016-10-05
Clarify magic indentation fixes applied to multi-line text edits in CompletionItem


kdvolder:
It looks like when I create a completion item which contains newlines, so expands to more than one line of actual text... vscode applies some kind of magic indentation fixes to my edit (rather than taking my edit at face value and replacing my text exactly as given).

This is disturbing really because my 'newText' already contains exactly the needed spaces to get the desired/correct indentation. I'm seeing that in vscode additional spaces are added on top of that which actually messes things up.

E.g my language server sends a completion like this:

```
  {
       "label":"services",
       "kind":5,
       "sortText":"00014",
       "filterText":"services",
       "textEdit": {
            "range":{
                  "start":{"line":2,"character":4},
                  "end":{"line":2,"character":4}
       },
      "newText":"services:\n      - "
 }
```

There are precisely 4 spaces between the `\n` and the `-`.

However, when the completion is invoked it looks like 8 spaces get added into the document instead.

Really, I'd prefer if 'newText' doesn't get subjected to this kind of magic stuff. But if there is indeed magic stuff happening, then we probably should have clear descriptions in the LS spec to explain exactly what magics to expect.

PS: This is somewhat related to #82 (both issues concern a different kind of undocumented 'magic interpretation' applied to the contents of the TextEdit's`newText` field).

dbaeumer:
@jrieken can you please comment. Is there a way to avoid the formatting (e.g. by using the new range property ?)

dbaeumer:
Moving to January. Need to first understand what exactly is happening and then need to think about a spec.

jrieken:
The *magic* is that VS Code prefixes new lines in snippets and completions with the [indentation](https://github.com/Microsoft/vscode/blob/master/src/vs/editor/contrib/snippet/common/snippet.ts#L100) of the 'reference line', that is the line on which the operation was started. So, if your reference line is indented with 4 spaces each new line in your snippet in prefixed with 4 spaces. This is in line with other editors and makes it easy to create snippets and completions without  always checking the indentation properties of the current document.

kdvolder:
Cool, okay. That clarifies it.

I think it it would be convenient if there was a way to enable/disable this magic, maybe by adding a boolean field 'enableMagicIndent' or something like that to the CompletionItem (or to the list to avoid repeating it for every item). To preserve backwards compatibility this field could be optional and the magic indent could be enabled by default if the field is absent.

kdvolder:
[...]

PR #221, opened 2017-04-19
Add a client capability for optional magic indents in completion items.

kdvolder:
This is a proposal to address the confusion around magic indentations, and make them optional, as discussed in #83

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -1845,7 +1845,7 @@ interface CompletionItem {
 	 */
 	insertTextFormat?: InsertTextFormat;
 	/**
-	 * Whether automatic relative indentation fixes are applied to a multi-line insert text.
+	 * Determines whether automatic relative indentation fixes are applied to a multi-line insert text.
 	 * This option applies to both the `insertText` property and the `newText` property
 	 * of a provided `textEdit`.
 	 *
@@ -1864,7 +1864,7 @@ interface CompletionItem {
 	 * To adjust indentation, the reference indent is then inserted into the insert text right
 	 * after every line break.
 	 */
-	relativeIndent?: boolean
+	disableRelativeIndent?: boolean
 	/**
 	 * An edit which is applied to a document when selecting this completion. When an edit is provided the value of
 	 * `insertText` is ignored.
--- a/protocol.md
+++ b/protocol.md
@@ -771,6 +771,15 @@ export interface TextDocumentClientCapabilities {
 		 */
 		completionItem?: {
 			/**
+			 * Client supports automatic indentation of multi-line insert text.
+			 *
+			 * A client that supports this provides the option for a multi-line insert text
+			 * (either snippet or plaintext) to have the indentation of all lines from the second
+			 * line onwards be adjusted relative the first line.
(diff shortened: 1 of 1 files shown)


protocol.md
```diff
@@ -1836,6 +1845,27 @@ interface CompletionItem {
...
+	 *
+	 * The reference indent is determined by taking all the whitespace
+	 * characters from the front of the reference line upto either the first non-whitespace
+	 * character; or the start position where the insert text is being applied. Whichever
+	 * of these two yields the shortest string is taken as the reference indent.
+	 *
+	 * To adjust indentation, the reference indent is then inserted into the insert text right
+	 * after every line break.
+	 */
+	relativeIndent?: boolean
```
svenefftinge: I'd assume that leaving this out will default to `relativeIndent = false`, but that is not the current behavior of vscode. So either we tunr the property name around (e.g. `disableRelativeIndent`) or it is stated that an undefined property means `true` (which sounds weird).

kdvolder: I don't feel strongly about it either way. However defaulting to not supporting magic indentations is probably what most people would assume if they read the current LSP description as it make no mention of these magic indents at all. Therefore, any implementor working from reading the spec would have no clue they are supposed to support these magic indents.

So in a way, I'd argue that vscode, in this case is actually not complying with the spec, as it is currently written.

kdvolder: Anyhow... I think you are right, it is probably slightly less disruptive to existing language servers if we flip the flag to be a 'disable' flag instead of a 'enable' flag. So I have changed that in my PR.

dbaeumer:
@kdvolder can you please outline why we need support to disable this per completion item. Is it to have completion items that unindent on the next line ?

We should if possible provide this as a proposed implementation as described here: https://github.com/Microsoft/language-server-protocol/blob/master/contributing.md

kdvolder:
> ...  why we need support to disable this per completion item. 

Actually, for our use-case we don't need it per item. It would be fine if we could disable it for all items (i.e. so everything from our language server). 

The reason why I proposed a 'per item' mechanism is that it felt like a similar thing to snippet support, so it felt logical to me to follow a similar mechanic for enabling / disabling it.

dbaeumer:
Understood. But this still needs to be implemented as a reference implementation. Simply adding the two booleans will not fix this for VS Code itself.

kdvolder:
I'm closing this PR. In light of recent discussions in https://github.com/Microsoft/language-server-protocol/issues/83 it seems rather to make more sense to avoid adding these kinds of complexities to the protocol and all clients that need to implement it. 

Instead it seems better if vscode should update its implementation to follow the spec as it is currently written (i.e. just take edits at face value) so that it behaves the same as the other clients.

Also to be honest, I probably do not have the bandwidth to carry through this PR with a reference implementation. It seems like a fair amount of work to me, and not worth spending this effort on supporting a vscode-specific behavior that most folks seem to consider is a bug, or at the very least an unwanted/undocumented vscode feature that accidentally filtered through into the vscode client implementation.

PR #231, opened 2017-05-03
Fix typo in summary of `interface Registration`.

CXuesong:
There seems to be an extra "to" in the annotation for the interface `Registration`.


dbaeumer:
Thanks.

Issue #70, opened 2016-09-19
Add support to report progress for a running request


dbaeumer:

felixfbecker:
Moving my suggestions here

I see two possibilities here
 - Go with the pattern of HTTP 202 Accepted. You send a request, and get back a response instantly stating the request has been "accepted". You can then do more requests with a different method, like `$/progess` (similar to cancel), passing the message ID of the initial request as a param, and get progress reports back.
 - Go with a pattern similar to HTTP 102 Processing (WebDAV). You send a request, and before the actual response comes in, the server can decide to continuously send a `Progress` header, and finally the actual `Content-Length` header and response body.

felixfbecker:
# Proposal

Add a new notification `window/progress` that is sent from the server to the client.
Params:
```ts
ProgressParams {
  /** a unique identifier to associate multiple progress notifications with the same progress */
  id: string | number;
  /** 
    * an optional request ID, indicating that this is the progress of a specific request from the client
    * if set, the client can use the information to present a progress indicator in the context of the action that triggered the request
    * if not set, the progress is treated as arbitrary progress for an arbitrary async operation (for example indexing) and the client can show it in a general location, like a status bar
    */
  requestId?: number | string;
  /** number that ranges from zero to total */
  status: number;
  /**
    * optional total number that status is relative to.
    * If set, the client can choose to display this next to the progress indicator.
    * If not set, assumed to be 1.
    */
  total?: number;
  /** optional label for the progress. If the client decides to use it, it should be displayed together with the progress status */
  label?: string;
}
```
It can be used to report progress for requests:
```
-> textDocument/references request (id 1)
<- window/progress notification (progress id 1, request id 1, status 0.5)
<- window/progress notification (progress id 1, request id 1, status 1)
<- textDocument/references response (id 1)
```
A progress indicator should be hidden by the client once a progress with status == total comes in.

It can also be used to report progress for file indexing for example:
```
<- window/progress notification (progress id 2, status 1, total 1264, label "indexing file hello.php")
<- window/progress notification (progress id 2, status 134, total 1264, label "indexing file world.php")
<- window/progress notification (progress id 2, status 1264, total 1264, label "indexing file whatever.php")
```

kaloyan-raev:
Should we have some feedback from the client to server if it was able to process a certain progress status?

[...]

PR #245, opened 2017-05-15
Add window/progress

keegancsmith:
VS Code 1.12 includes an API for progress:

```ts
vscode.window.showProgress({
    location: vscode.ProgressLocation.Window,
    title: 'My long running operation'
}, async progress => {
    // Progress is shown while this function runs.
    // It can also return a promise which is then awaited
    progress.report({ message: 'Doing this' });
    await step1();

    progress.report({ message: 'Doing that' });
    await step2();
})
```

This proposal intends to make it possible for LSP to provide similar functionality.

The progress extension in vscode allows logging to source control and the window. However, we do not include that ability in this spec since:

* It only seems appropriate for LSP to log to the window
* The location changes what sort of thing you can log https://sourcegraph.com/github.com/Microsoft/vscode@1.12.1/-/blob/src/vs/workbench/api/node/extHost.api.impl.ts#L356

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -36,6 +36,7 @@ Window
 * :arrow_left: [window/showMessage](#window_showMessage)
 * :arrow_right_hook: [window/showMessageRequest](#window_showMessageRequest)
 * :arrow_left: [window/logMessage](#window_logMessage)
+* **New** :arrow_left: [window/progress](#window_progress)
 * :arrow_left: [telemetry/event](#telemetry_event)
 
 >**New** Client
@@ -611,7 +612,7 @@ The initialize request is sent as the first request from the client to the serve
 
 Until the server has responded to the `initialize` request with an `InitializeResult` the client must not sent any additional requests or notifications to the server. 
 
->**Updated**: During the `initialize` request the server is allowed to sent the notifications `window/showMessage`, `window/logMessage` and `telemetry/event` as well as the `window/showMessageRequest` request to the client.
+>**Updated**: During the `initialize` request the server is allowed to sent the notifications `window/showMessage`, `window/logMessage`, `window/progress` and `telemetry/event` as well as the `window/showMessageRequest` request to the client.
 
 _Request_:
 * method: 'initialize'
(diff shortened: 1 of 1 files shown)


dbaeumer:
@keegancsmith thanks for the PR. Actually adding this needs to be done under considering how other tools might implement something like this. The LSP client side is implemented in Eclipse, GNome builder, ...

I will mark this one as a dup of https://github.com/Microsoft/language-server-protocol/issues/70 which already contains quite some discussions around this topic.

felixfbecker:
@dbaeumer Is there anything specific here that would not work with other editors? This follows the ideas in #70 pretty closely

Xanewok:
@felixfbecker Do you think we could introduce this simple notification for starters, see how it goes and refine from there? I believe standarizing a simple progress protocol would incentivize more clients to implement at least some form of reporting the progress.
This could give more useful feedback on the issue, I think.

Currently the Rust Language Server uses custom `rustDocument/diagnostics{Begin,End}` notifications to report analysis build in progress and places a burden on the client to keep the 'running' diagnostics counter - total count of 0 means the build is done and all the information is up to date.
We were considering adding a bit more information (e.g. which dep is built, in addition to the main workspace package(s)), but as I said, right now it seems a bit ad hoc and we're mostly waiting for the LSP to standarize the feature.

felixfbecker:
@Xanewok lets move discussion to #70

felixfbecker:
Sure, I think if multiple servers or clients want this it is reasonable to implement as an experimental/extension method and see if it picks up, then repropose it again. The way we did this with extensions at Sourcegraph was to prefix the methods with `x`, i.e. `window/xprogress`. Of course it's only valuable if we get at least one client to implement it too. 

Is the PR in it's current form sufficient for your needs? Any other suggestions?

Xanewok:
Regarding the current form, I think a more practical/consistent approach would for the
```diff
/**
-   * The title of the progress.
-   * This should be the same for all ProgressParams with the same id.
+   * Optional title of the progress.
+   * If unset, the previous title (if any) is still valid.
 */
-  title: string;
+  title?: string;
```

Seems redundant and/or hard to enforce passing the same title for a given id, but with the above change it'd be possible not to set any without introducing a new message - something I'm not sure is desired.

I think it's simple and self-contained enough in the current form as-is, I think it'd be nice to introduce it as an initial/experimental version.

@dbaeumer what are your thoughts?

felixfbecker:
That makes a lot of sense to me. Wanna do a PR to the sourcegraph repo? We can have this live there as an unofficial spec extension for the time being

dbaeumer:
I added my comments here:  https://github.com/Microsoft/vscode-languageserver-node/pull/261#issuecomment-336072641. I am open to where the discussion should happen but since a lively discussion happened in the PR I thought it is a good place to have it.

PR #284, opened 2017-08-10
Multi Line CompletionItem (fixes #283)

### LINKED ISSUE #283: Clarify how to deal with indentation on multiline text inserts
When applying a completion item, with a multi line snippet or text to insert. The behavior of monaco is to append the current indentation to newlines.
I.e.
```
class Foo {
   <CURSOR>
}
```
where the text to insert is e.g. 
```
function foo() {
}
```

I get
```
class Foo {
   function foo() {
   }
}
```

This behavior should be added to the protocol to clarify the expectations on client and server side.

svenefftinge:
Added clarification about multi line text to insert and who is responsible for adding contextual indentation.

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -1827,8 +1827,11 @@ interface CompletionItem {
 	 */
 	filterText?: string;
 	/**
-	 * A string that should be inserted a document when selecting
+	 * A string that should be inserted into a document when selecting
 	 * this completion. When `falsy` the label is used.
+	 * 
+	 * The string might contain line breaks, in which case it is the client's responsibility to prepend new lines with 
+	 * the current indentation.
 	 */
 	insertText?: string;
 	/**


mickaelistria:
Why should the complexity here be passed to the client when the server has all the necessary info to just include the right identation?

kdvolder:
The wording is a little too vague I think. It may seem 'obvious' what 'current indentation' means to you but I don't think it is.

Also I agree with @mickaelistria that its not more of an annoyance than a help in many situations when the client performs 'magic' indentations when the server is capable of generating snippets that are already perfectly spaced.

See a PR and issue that I raised a while ago about this: 

- https://github.com/Microsoft/language-server-protocol/pull/221
- https://github.com/Microsoft/language-server-protocol/issues/83

svenefftinge:
> Why should the complexity here be passed to the client when the server has all the necessary info to just include the right identation?
Because VS Code does it so.

svenefftinge:
closed in favor of #221

PR #315, opened 2017-10-08
Fix links and mention used `protocol.` prefix

Xanewok:


dbaeumer:
@Xanewok the path separator got already fixed. I merge the work smithing by hand.

PR #326, opened 2017-10-28
Moved the last details to a better place

CWBudde:
So far 'Applies a WorkspaceEdit' and 'Execute a command' was located at the end. This commit aligns both with the other workspace details.


dbaeumer:
Merged by hand since outdated and moved some more parts as well.

PR #328, opened 2017-11-01
Fix links to protocol source

dstjacques:


dbaeumer:
@dstjacques thanks. I merged by hand since the document already evolved.

PR #331, opened 2017-11-09
Specify response from client/unregisterCapability

vladdu:
Assuming that it is void, just like for registerCapability.


dbaeumer:
Yes, it is void.

PR #337, opened 2017-11-16
Add ClientCapabilities.dynamicRegistration

mickaelistria:
The documentation about client/registerCapability references a ClientCapabilities.dynamicRegistration field that's missing from the current ClientCapabilities description (so it's easily forgotten in related libraries such as LSP4J).

### DIFF
--- a/protocol.md
+++ b/protocol.md
@@ -960,6 +960,11 @@ interface ClientCapabilities {
 	textDocument?: TextDocumentClientCapabilities;
 
 	/**
+	 * Client supports the `client/registerCapability` request from server.
+	 */
+	dynamicRegistration?: Boolean;
+
+	/**
 	 * Experimental client capabilities.
 	 */
 	experimental?: any;


dbaeumer:
The `dynamicRegistration` capability is not global. It is per feature since it might not be feasible for all features for a client. See `TextDocumentClientCapabilities`. Objections to close?

mickaelistria:
Before we close, I need a clarification then: which one of the existing capabilities is referenced by
> Not all clients need to support dynamic capability registration. A client opts in via the ClientCapabilities.dynamicRegistration property.

As here we speak about dynamic registration of capabilities, it doesn't seem right to include such field in a dedicated service (Workspace or TextDocument), because this capability is server-wide.
Note that a solution to this confusion in the protocol can also be to remove the "Not all clients need to support dynamic capability registration. A client opts in via the ClientCapabilities.dynamicRegistration property." block and assume all clients need to support dynamic registration, but that wouldn't be backward-compatible.

dbaeumer:
This is made optional especially because of backwards compatibility and not all clients have the ability to ad for example a code complete provider when a file is already open. So there needs to be some upfront registration. 

I update the text to 

> A client opts in via the `dynamicRegistration` property on the specific client capabilities. A client can even provide dynamic registration for capability A but not for capability B (see `TextDocumentClientCapabilities` and an example).

to make this more clear.

mickaelistria:
Ok, I get it.
I was actually confused by the usage JDT-LS does of this dynamicRegistration for workspaceFolders. As there is no existing capability, I think it's logically in an inconsistent state according to the spec so it sends registerCapabilities anyway.

dbaeumer:
It does. Since the workspace folders are still proposed in is in the proposed spec part :-)

PR #364, opened 2017-12-29
Add design polish to public pages

auchenberg:
- New logo
- Add new /history section
- Mobile optimization
- Better spacing
- Various tweaks

screen shot 2017-12-29 at 9 27 39 am


css/main.scss
```diff
@@ -9,16 +9,38 @@ html {
 }
 
 body {
-	/* we have a top fixed nav bar */
-	margin-top: 50px;
-	margin-bottom: 46px;
+	letter-spacing: .25px;
+    text-rendering: optimizeLegibility;
```
rictic: Nit: mixture of tabs and spaces here, better to pick one.

overview.md
```diff
@@ -1,9 +1,18 @@
 ---
-title: How it Works
+title: Overview
 layout: singlePage
-sectionid: howItWorks
+sectionid: overview
 ---
 
+## What is the Language Server Protocol?
+Implementing language support like auto complete, goto definition, or type hovering for a programming language for different development tools is a significant effort. Each development tool provides different APIs for language extenders. Moreover, an extension has to be implemented in the implementation language of the tool itself.
```
rictic: Rewording and flow suggestions. Take what parts you like:

> Implementing support for features like autocomplete, goto definition, or documentation on hover for a programming language is a significant effort. Traditionally this work must be repeated for each development tool, as each provides different APIs for implementing the same features.

> The idea behind a <i>Language Server</i> is to provide the language-specific smarts inside a server that can communicate with development tooling over inter-process communication. The idea behind the <i>Language Server Protocol</i> is to standardize how tools and servers communicate, so that a single language server can talk to many different development tools, and a single tool can talk to many language servers!

css/main.scss
auchenberg: Good catch. Fixed.

index.html
```diff
@@ -2,73 +2,76 @@
...
-              </p>
-            </blockquote>
+
+            <p>Implementing support for features like autocomplete, goto definition, or documentation on hover for a programming language is a significant effort. Traditionally this work must be repeated for each development tool, as each provides different APIs for implementing the same features.</p>
+            <p>The idea behind a Language Server is to provide the language-specific smarts inside a server that can communicate with development tooling inter-process communication through a protocol.</p>
```
rictic: Looks like a typo or missing word in: "with development tooling inter-process communication"

Maybe `via inter-process communication`?

rictic: Ending this sentence with `through a protocol` is a tad unnecessary given that the next sentence introduces the protocol, and since all IPC communication is necessarily through a protocol, if perhaps an ad-hoc one.

auchenberg:
@dbaeumer @egamma Let me know what you think, and whether we for example should get these tweaks in without the proposed logo.

dbaeumer:
@auchenberg thanks a lot for doing this work. I looked over the changes and here are some comments/questions:

- I would personally go without the logo. The best I could think of is having an acronym (LSP) in the form of the Bootstrap logo. It will be very hard to come up with a graphic that conveys the meaning of a language server protocol. 
- I tried to mimic the design we have for code.visualstudio.com so I would like to understand why you did some of the changes:
   - removed the fix navigation bar. I actually liked it :-)
   - the left navigation of new pages (e.g. Implementation) and the right navigation inside a page (specification). I know that the specification page is empty on the left right now but I think we should consider breaking the huge specification page up anyways. Then we would have a left and right navigation for that page.
- I like the footer but it looks a little empty. Should we steal some ideas from code.visualstudio.com.
- the history page makes sense but without more content I would leave it out for now.

auchenberg:
@dbaeumer 

1. I agree. Until we have a proper brand we shouldn't have a pseudo solution, so let's go with the LSP / Language Server Protocol naming. Logo removed.

2. I removed the fixed topbar as the side navigation is fixed/floating and so to say takes priority of the top navigation when you scroll, as the navigation then is about navigating the page content. We do the same on code.visualstudio.com

3. I'd buy that argument. I'll bring it back

4. We should. Let me copy/paste :) Just wanted to sneak a Microsoft logo in there in tasteful way.

5. Indeed. This page would need some content before going online. Mostly thought as a place for Erich to write-up the history and link to the press announcements, etc. More links, better SEO.

dbaeumer:
@auchenberg:

2. : now I see. The side bar scroll when you reach the bottom of the page. I have to say that I still a fixed navigation bar. Bootstrap web page solved this differently by making the side navigation bars scrollable. Would like be an option for you ? (http://getbootstrap.com/docs/4.0/getting-started/introduction/)

auchenberg:
@dbaeumer Please have a look. 

- Fixed top level menues are back. 
- Side navs are sticky
- New mobile side nav
- Right nav for inpage content
- Logo removed
- New favicon
- Responsive tables
- More content in footer

What's left is more content on `/history`

auchenberg:
Ready to go @dbaeumer ?

auchenberg:
Taking the action to merge this in. Should be up to date with recent changes in `gh-pages` branch.

dbaeumer:
@auchenberg thanks for doing this and merging this in. I am busy with the language pack work and ignored issues for some days.

PR #373, opened 2018-01-12
Change message params types to match JSON-RPC spec

Xanewok:
Since the interface is supposed to describe message format to match [JSON-RPC spec](http://www.jsonrpc.org/specification),
the params type must be specified as either `Array` or `Object`.

From spec:
> 4.2 Parameter Structures
>
> If present, parameters for the rpc call MUST be provided as a Structured value. Either by-position through an Array or by-name through an Object.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -195,6 +195,8 @@ A request that got canceled still needs to return from the server and send a res
 
 The language server protocol defines a set of JSON-RPC request, response and notification messages which are exchanged using the above base protocol. This section starts describing the basic JSON structures used in the protocol. The document uses TypeScript interfaces to describe these. Based on the basic JSON structures, the actual requests with their responses and the notifications are described.
 
+In general, the language server protocol supports JSON-RPC messages, however the base protocol defined here uses a convention such that the parameters passed to request/notification messages should be of `object` type (if passed at all). However, this does not disallow using `Array` parameter types in custom messages.
+
 The protocol currently assumes that one server serves one tool. There is currently no support in the protocol to share one server between different tools. Such a sharing would require additional protocol to either lock a document to support concurrent editing.
 
 ### Basic JSON Structures
--- a/specification.md
+++ b/specification.md
@@ -89,7 +89,7 @@ interface RequestMessage extends Message {
 	/**
 	 * The method's params.
 	 */
-	params?: any
+	params?: Array<T> | object;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Xanewok:
Analogous to https://github.com/jacobdufault/cquery/issues/247, sending not structured values as params (which is currently allowed here via `any` type), if present, can upset brittle clients/servers, which are parsing messages literally as per JSON-RPC spec.

dbaeumer:
Although we use JSON-RPC as a transport layer we decided to limit the params to a single object. This is why on the specification level we ony support a single object. From a spec point of view we should change the property to `object`. Does that make sense.

Xanewok:
I understand that such a limitation might be convenient, however I believe there's a large benefit to retaining the compositional nature of the protocol. It's beneficial to be able to reuse existing JSON-RPC libs and format as is - we could simply add a note, such as 'Array params are technically allowed, but practically unused in LSP' to maintain the compatibility with the JSON-RPC protocol. This would be a win-win, I believe.

dbaeumer:
We do so in the json-rpc library we use and maintain. It allows to handle arrays correct according to the JSON-RPC spec. Limiting this in the protocol is IMO a fair thing to do and nothing we can change without starting to break since users can implement custom messages. I don't see why we should allow array parameters and then say that they very likely will not work. Implementors of the protocol can still use existing JSON-RPC libs since they will work as design with 0 / 1 parameter.

Xanewok:
I realize this is mostly nitpicking, but we should be more consistent here: we can't just say we use and define interface for JSON-RPC messages, yet do otherwise.
> Limiting this in the protocol is IMO a fair thing to do and nothing we can change without starting to break since users can implement custom messages.

The thing is, someone could've implemented custom messages passing params as arrays, which is correct under current phrasing that we use JSON-RPC spec, and also currently (in contradiction to previous phrasing) specified `any` type for params.

If we were to limit legal params type only to `object`, then we both technically break JSON-RPC compliance and break possible custom messages.

That is why I meant that we could still allow for `Array<any>` params - this would still leave some leeway for other possible custom messages to still possibly use legal array params as per JSON-RPC spec and would also disallow other illegal params types (e.g. primitives).

Xanewok:
@dbaeumer done, I mentioned in the section that `object` types here are only used, but in general the LSP shouldn't disallow params of `Array<any>` type.

PR #381, opened 2018-01-19
Add cquery as an implementation of LSP for C/C++/Objective-C

khng300:
CQuery is an LSP implementation based on clang frontend. If you wish to know more about its functionalities, you may visit https://github.com/jacobdufault/cquery/blob/master/README.md .


dbaeumer:
Merged in by hand.

PR #410, opened 2018-02-21
Fix incorrect OmniSharp link

agc93:
Existing link for OmniSharp server was directing to the node client repo


egamma:
@agc93 thank you!

Issue #54, opened 2016-08-23
Clarification for the Indexing workflow


kaloyan-raev:
Many languages require building an index database for an efficient calculation of code completion, validation, highlighting, etc.

The language server protocol currently does not describe how the communication between the client and the server should happen. Thus, every server-client pair take their own approach. For example the [Crane PHP language server](https://github.com/HvyIndustries/crane) uses custom message requests and notifications, while the [Eclipse Java language server](https://github.com/gorkem/java-language-server) seems to trigger the indexing as part of the initialize request.

Please, consider specifying:
- How the client request the server to (re-)build the index.
- How the client tells the server which files are included and excluded for indexing.
- How the server reports progress to the client.
- How the server notifies the client that the index is complete.
- How the server notifies the client for any errors.

felixfbecker:
Isn't this covered by `didChange`, `didChangeWatchedFiles` etc? Should the client really care about how the server implements the index?

kaloyan-raev:
I guess some existing part of the protocol can be used.

Let's take as example the question "When the language server should start indexing the project?". A possible answer is "The indexing should be triggered when the client sends the `initialize` request.". So far so good. It still needs to be documented, so all clients and servers know this is the way to do it.

But, what if the client needs to send the server some configuration related to the indexing. For example, some clients may want to include/exclude some folders from the indexer. The `didChangeConfiguration` notification goes after the `initialize` request, which is late (the indexing is already going). VSCode has an additional `initializationOptions` as part of `InitializeParams`, but it is not documented in the protocol. See #55.

How should the settings related to the indexing should look like? There should be some JSON object defined.

Then how the server should report the indexing progress to the client? Via `logMessage` or `showMessage` notifications? Or something else?

How should the client request a full index rebuild? I don't find an appropriate existing request/notification in the protocol.

felixfbecker:
[...]

PR #438, opened 2018-03-30
Send configuration on language server startup

arxanas:
There's currently a method of sending configuration to the language server using the `workspace/didChangeConfiguration` notification. However, the configuration is not available on language server initialization. This could be benign, or it could mean that the language server has to stop doing work with a default configuration and start doing work with the updated configuration, such as in #54.

In the Hack language server, we would like to support a "dynamic view", which shows additional type errors for the file that you're looking at using a different, more strict interpretation of the `any` type (cc @jamesjwu). This can be toggled on and off in the editor as the user needs it. It's not feasible to restart the language server whenever we want to change this setting because the language server takes anywhere from 30 seconds to 10 minutes to start up.

The difference between `initializationOptions` and `settings`, then, is that the former doesn't change once the language server is initialized, and the latter can change when the user changes their configuration in the editor.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -723,6 +723,12 @@ interface InitializeParams {
 	 */
 	initializationOptions?: any;
 
+	/*
+	 * Configuration settings in the client. These can be changed later by the
+	 * client by issuing a `workspace/didChangeConfiguration` notification.
+	 */
+	settings?: any;
+
 	/**
 	 * The capabilities provided by the client (editor or tool)
 	 */


rcjsuen:
I noticed that the content of the opening comment is exactly the same as the commit message of a8fd03c9588d1b4738b6bcd2bce3aa583346c115. While this is usually not a problem, I am not sure words like "cc Somebody" belong in a commit message...?

dbaeumer:
I am actually against adding the settings to the initialize request for the following reasons:

- there are new request and notifcation that a server can sent to the client to access configuration settings in an improved way. So the old configuration push model got basically replaced by a new pull model.
- there is since a while a `initialized` notification sent from the client to the server when the client is ready to receive request and notifications. A server can use this notification to pull configuration from the client and setup itself correctly.

@arxanas would that model work for you. If not please let me know what is missing. I am more in favor of improving the new setting model (pull) than the old one.

arxanas:
@dbaeumer: That model seems fine. However, for our purposes, the old configuration push model can't be completely replaced by the new pull model. In our case, we won't have any idea when the user has changed their configuration (by selecting a button in the editor to toggle it), so we would have to poll the client if we couldn't listen for a configuration-change event.

dbaeumer:
@arxanas the VS Code LSP client still sends configuration changed notification even when the new pull model is used. However the notification is empty (no settings). Which client are you using?

arxanas:
@dbaeumer: We're using Nuclide. I don't think we have the configuration requests implemented currently, so we'll have to implement these two methods.

PR #441, opened 2018-04-04
Change workspace/configuration's icon to be a request

rcjsuen:
`workspace/configuration` is a request so it should use a hook instead of an arrow to visually indicate to the reader that something will be sent back to the originator of the message.


dbaeumer:
Thanks !

PR #444, opened 2018-04-05
Fix for sticky sidebar going off page on Desktop

caleblloyd:
Presently the sticky sidebar is off the screen when expanded on desktop.  It is not possible to scroll the sticky sidebar.  I have a 1080p screen so this affects many users.  Solution in this PR is to disable stickiness on the sidebar.

screenshot-2018-4-5 specification


caleblloyd:
The sticky sidebar is a nice feature when browsing long docs.  I have updated the PR to add scrolling to the sticky sidebar when it is taller than the screen with `overflow: auto`

auchenberg:
LGTM!

PR #448, opened 2018-04-11
Fix #447 Clarify the default for TextDocumentSyncOptions's change

### LINKED ISSUE #447: What is the default value for ServerCapabilities.textDocumentSync and TextDocumentSyncOptions.change
It isn't clear from the spec, but what is the default behaviour if these optional fields are not specified? One would suspect in both cases it is equivalent to `TextDocumentSyncKind.None` when not specified.

rcjsuen:
a1b13c56810a2617fcfff9c280d4098602a4fb2f didn't include a change to `TextDocumentSyncOptions`. This pull request adds the text to `TextDocumentSyncOptions`'s `change` field also.


dbaeumer:
@rcjsuen thanks. Good catch.

PR #476, opened 2018-05-09
change ruby implementor to solargraph

dre-hh:
There was not much activity on language_server-ruby recently. 
It was using rcodetools (a very outdated gem, which could not work properly)
The gem was removed. The only ruby native solution for static analysis available is solargraph
And it has implemented a language server on its own, which is quite feature complete.
The language server was also used by sourcegraph.
https://github.com/castwide/solargraph/issues/8


dbaeumer:
@dre-hh I merged in the changes by hand to keep the old reference as well. Hope that is OK for you.

Issue #389, opened 2018-01-25
Add CodeAction class 


mjbvz:
VS Code 1.20 adds a new `CodeAction` type that captures additional information about codeactions:

- https://github.com/Microsoft/vscode/blob/b5e2d0a6bcbae4f71099715dc813239d21975eb5/src/vs/vscode.d.ts#L1897
- https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_20.md#codeactionprovider-improvements

This new type is used by several new product features such as the `Refactor` command and the `editor.action.codeAction` command. It should also be supported in the LSP for code action provides

gorkem:
+1 
I think the `WorkspaceEdit` field is also an improvement on the references object. Most code actions are just workspace edits and some LSPs are doing a round trip for no good reason other than specification's short comings.

dbaeumer:
See also the discussion in https://github.com/Microsoft/language-server-protocol/issues/432. We should work on this sooner than later since it will make things a lot easier to understand when it comes to commands and code actions.

rcjsuen:
New `OrganizeImports` `CodeActionKind` being considered, see Microsoft/vscode#47845.

mjbvz:
I just merged in the `OrganizeImports` addition from Microsoft/vscode#47845. From the API perspective, all it did was add a new constant.

I'd like to get other extensions that have their own organize imports (such as java) to try to migrate over to using a code action instead. This is blocked until we have basic support for a `CodeAction` class in the LSP

mickaelistria:
I would like to work on this, however I've got some preliminary questions:
Does the proposal for the spec have to follow the current class in VSCode? I'm not convinced we need to specify the codeActionKind from 1st round, and I also find more correct for the protocol to clearly use `Command|WorkspaceEdit` explicitly instead of having 2 distinct field and a comment saying they're exclusive.
Also, should the protocol add some client/server capability to declare what kind of codeActions results are expected/supported (current command-only as default value, codeAction class as an option)? Or can we just move it forward and let clients implement some tricks for backward compatibility (LSP4J is good at it as it can have the request return an `Either<Command[]|CodeAction[]>` for the codeAction request, which is IMO simpler to use than capabilities to decide of how to handle the response).

dbaeumer:
Actually this got added to the VS Code May plan :-). 

[...]

PR #534, opened 2018-07-28
Add CodeActionOptions with providedCodeActionKinds

Gama11:
See here for the motivation: https://github.com/Microsoft/language-server-protocol/issues/389#issuecomment-407717530

Corresponding vscode-languageclient PR: https://github.com/Microsoft/vscode-languageserver-node/pull/385

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1327,6 +1327,19 @@ export interface SignatureHelpOptions {
 }
 
 /**
+ * Code Action options.
+ */
+export interface CodeActionOptions {
+	/**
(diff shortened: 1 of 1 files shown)


dbaeumer:
Adding this makes complete sense to me.

dbaeumer:
I renamed the property to `codeActionKinds` to make it consistent with the LSP naming. We never have `provided` in the capability name if the server provides something. The provide is already part of the top level naming (e.g. `codeActionProvider`)

PR #538, opened 2018-08-06
Add hierarchicalDocumentSymbolSupport to Text Document Client Capabilities

Bubba:
This is already returned by the VS Code insiders build and in https://github.com/Microsoft/vscode-languageserver-node/blob/451eee78181d529cfa98ffc7198e2955cdfdc451/protocol/src/protocol.ts#L425.
It seems necessary in order to maintain backwards compatibility with clients that do not yet support the new `DocumentSymbol` type.


aeschli:
Thanks @Bubba !

PR #546, opened 2018-08-19
Add lsp-test to SDKs

Bubba:
It's not really an SDK but a framework for writing tests against LSP servers


dbaeumer:
Thanks. Still worthwhile to be listed.

PR #548, opened 2018-08-21
Don't require the detail to be copy of name

matklad:
cc https://github.com/Microsoft/vscode-languageserver-node/commit/6c3f8d6891ac651e2b7f9dee4e012fa75be2a5d2, @aeschli


aeschli:
Thanks, right, I forgot that.

PR #549, opened 2018-08-22
Fixed textDocument/foldingRanges => textDocument/foldingRange

spoenemann:
This was already changed in most places, but one remains.


aeschli:
Thanks!

PR #550, opened 2018-08-22
Renamed FoldingRangeRequestParam => FoldingRangeParams

spoenemann:
For all other requests, the single parameter type is named `...Params`. For consistency, this pattern should also apply to the new folding range request.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -3876,10 +3876,10 @@ The folding range request is sent from the client to the server to return all fo
 _Request_:
 
 * method: 'textDocument/foldingRange'
-* params: `FoldingRangeRequestParam` defined as follows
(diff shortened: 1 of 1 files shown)


spoenemann:
By the way, the `export` keyword doesn't make any sense in the specification. Should it be stripped off?

rcjsuen:
Personally, I would prefer changing it to `FoldingRangeParams` and dropping `Request` from the name. Outside of `ShowMessageRequestParams`, all the other parameters do not include the word `Request` in its interface name.

dbaeumer:
+1 for removing `request`

spoenemann:
Done.

dbaeumer:
@aeschli can you please decide on this. We should do this sooner than later

PR #572, opened 2018-09-23
Allow MarkupContent[] to be used as Hover contents

### LINKED ISSUE #518: MarkupContent in Hover lacks support for "regions"
(as discussed in https://github.com/Microsoft/vscode-languageserver-node/issues/328 and https://github.com/Microsoft/vscode-languageserver-node/issues/374)

When using `MarkedString[]` as a Hover's `contents`, a client can interpret this as different "regions" and format it accordingly, e.g. with a horizontal separator like VSCode does. However, `MarkedString` is deprecated, and this can't be communicated with `MarkupContent`. So effectively, one is forced to use deprecated API to achieve this formatting right now.

Perhaps `MarkupContent[]` could be allowed as a possible type for `Hover.contents` as well to remedy this.

DJMcNab:
Fixes #518 and Microsoft/vscode-languageserver-node#374.

Supported by Microsoft/vscode-languageserver-node#417

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -3038,7 +3038,7 @@ interface Hover {
 	/**
 	 * The hover's content
 	 */
-	contents: MarkedString | MarkedString[] | MarkupContent;
+	contents: MarkedString | MarkedString[] | MarkupContent | MarkupContent[];
 
 	/**
 	 * An optional range is a range inside a text document


DJMcNab:
Closing after discussion with @dbaeumer in Microsoft/vscode-languageserver-node#417. Note that `atom-languageclient` does not use this same approach (at least when I last checked.

Sourcegraph uses css rules to make the markdown content with horizontal seperators look the same as multiple array items: <https://github.com/sourcegraph/codeintellify/commit/2ab2bf62289fc0266b3772fe07aee9d82c1786e7>

PR #575, opened 2018-09-28
overview: Fix example response id

Turbo87:
The `id` field in JSON-RPC is numeric (matching the type in the request)


dbaeumer:
Correct. Thanks for the PR.

PR #576, opened 2018-10-02
Add server capability for potential commit characters

### LINKED ISSUE #527: Allow server to announce potential commit characters
Currently commit characters are sent per `CompletionItem`. Some clients may require a set of "potential" commit characters upfront for performance reasons. There should be a field on `ServerCapabilities` that announces which characters may be used to commit a `CompletionItem`

ZoeyR:
Closes #527 

This PR adds a capability to the server that advertises the potential commit characters that can be used. This is useful so that clients can optimize their typing code by limiting the set of characters that could potentially cause a commit.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1531,6 +1531,11 @@ export interface CompletionOptions {
 	 * The characters that trigger completion automatically.
 	 */
 	triggerCharacters?: string[];
+
+	/**
+	 * The list of all possible completion commit characters
+	 */
+	potentialCommitCharacters?: string[];
 }
 /**
  * Signature help options.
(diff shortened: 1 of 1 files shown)


dbaeumer:
Sorry for the long delay. I would suggest we name it allCommitCharacters. `Potential` sounds strange to me.

dgriffen:
I'm fine with changing the name to `allCommitCharacters`, `potentialCommitCharacters` was just the terminology used by the VS completion APIs.

dbaeumer:
I added it to the node modules. Will update the spec when we release 3.15

dbaeumer:
Got obsolete. The spec is updated.

PR #593, opened 2018-11-05
Add Nim to the list of language identifiers

PMunch:
Now that the Nim programming language has an early stages, but working, language server I think it's time to add it to the list of identifiers.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -643,6 +643,7 @@ Less | `less`
 Lua | `lua`
 Makefile | `makefile`
 Markdown | `markdown`
+Nim | `nim`
 Objective-C | `objective-c`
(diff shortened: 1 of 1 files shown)


dbaeumer:
@PMunch there is a difference between listing a server on the server page and adding a new language to the list of well know language ids. I have no problem of adding Nim to this list https://microsoft.github.io/language-server-protocol/implementors/servers/ (PR necessary against the corresponding servers.md file). But listing it as a well know standard language ID usually requires a broader use of that language in the community.

Please note that you can still use the `nim` langueg id.

dbaeumer:
@PMunch. Thanks for pointing this out. I created: https://github.com/Microsoft/language-server-protocol/issues/598

PR #601, opened 2018-11-14
Add implementation of Swift to LSP list

unnamedd:
Apple released it's own implementation of LSP, here I'm suggesting to add it on the list.


rcjsuen:
What a time to be alive...

Issue #566, opened 2018-09-08
Error handling in Rename


doriath:
I have a question about properly handling errors in `textDocument/rename`. I wonder how server should respond to following scenarios:

1. When `textDocument` or `position` in `RenameParams` in invalid (not existing file, position out of bound, etc.) 
1. When there is a compile error in the project, and we are unable to actually perform any renames
1. When the `position` points to a symbol that can't be renamed (e.g. we point to an space character)
1. What error code should be used if `newName` is invalid?

Also, the server returns result: `WorkspaceEdit | null` - what is the valid use of `null`? Is it when there is nothing to change, or for some of the cases above?

Would it be useful to clarify those in the actual specification, maybe giving some examples?

LaurentTreguier:
For 1. I would send a response with an `InvalidParams` error instead of a result, because there is an error in the client's params (although technically it could be a bug in the server's internal document synchronization).
2. and 3. are not client errors, so I would send back a `null` result.
Same for `newName` being invalid, it's also not a request error (I'm assuming that it's invalid according to the language spec).

As far as I understand, sending `null` serves to indicate that a symbol can't be renamed for one reason or another.

dbaeumer:
There is now also a prepareRename request. This is send before the user is actually asked for a new name. It allows to specify the range to be renamed, a new name proposal.

dbaeumer:
I am all in for clarifying the spec here. As always PRs are welcome.

PR #616, opened 2018-11-28
Clarify responses to rename request.

doriath:
The main goal is to clarify how the `null` value should be used.
Currently it is not clear in what cases server is allowed to return it,
and how clients should interpret it. #566

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -4063,8 +4063,8 @@ interface RenameParams {
 ```
 
 _Response_:
-* result: [`WorkspaceEdit`](#workspaceedit) \| `null` describing the modification to the workspace.
-* error: code and message set in case an exception happens during the rename request.
+* result: [`WorkspaceEdit`](#workspaceedit) \| `null` describing the modification to the workspace. `null` should be treated the same was as [`WorkspaceEdit`](#workspaceedit) with no changes, and should only be returned when `newName` is the same as the old name (no change was required).
+* error: code and message set in case when rename could not be performed for any reason. Examples include: there is nothing at given `position` to rename (like a space), given symbol does not support renaming by the server or the code is invalid (e.g. does not compile).
 
 _Registration Options_: `RenameRegistrationOptions` defined as follows:
 


doriath:
### Why?

First, let me explain why I would like to add this clarification here. ~2 months ago I was playing with RLS (Rust language server) and vim-lsp (client for vim) and I found that rename was not a great user experience.

Here is why:
- RLS returns `null` when rename did not succeed (for any reason). This also included cases like the code doesn't compile, so rename cannot be performed
- vim-lsp does nothing when `null` is returned

This means that when I tried to rename something and it failed, I had no message telling me that something failed. From the client (vim-lsp) perspective, it is not clear how to interpret the `null` response - was there just nothing to do, or was there some error and rename was not performed.

### Alternative

In my Pull Request, I went with the proposal of treating `null` as "nothing has to be done". The alternative approach could be to use it as generic "rename failed" indication.

If this one is preferred, I would propose:
- clarify that clients can treat this value as "can't perform rename at this position"
- generally discourage server implementations from returning null, in favor of returning clear error message. For example: the position is not at any symbol (e.g. space or brackets), current symbol does not support renames or the code doesn't compile.

After writing this explanation, I am slightly starting to lean toward the alternative proposal, but I will wait for the feedback first.

specification.md
```diff
@@ -4063,8 +4063,8 @@ interface RenameParams {
 ```
 
 _Response_:
-* result: [`WorkspaceEdit`](#workspaceedit) \| `null` describing the modification to the workspace.
-* error: code and message set in case an exception happens during the rename request.
+* result: [`WorkspaceEdit`](#workspaceedit) \| `null` describing the modification to the workspace. `null` should be treated the same was as [`WorkspaceEdit`](#workspaceedit) with no changes, and should only be returned when `newName` is the same as the old name (no change was required).
```
KamasamaK: Should be "the same _way_ as".

specification.md
```diff
@@ -4063,8 +4063,8 @@ interface RenameParams {
 ```
 
 _Response_:
-* result: [`WorkspaceEdit`](#workspaceedit) \| `null` describing the modification to the workspace.
-* error: code and message set in case an exception happens during the rename request.
+* result: [`WorkspaceEdit`](#workspaceedit) \| `null` describing the modification to the workspace. `null` should be treated the same was as [`WorkspaceEdit`](#workspaceedit) with no changes, and should only be returned when `newName` is the same as the old name (no change was required).
+* error: code and message set in case when rename could not be performed for any reason. Examples include: there is nothing at given `position` to rename (like a space), given symbol does not support renaming by the server or the code is invalid (e.g. does not compile).
```
KamasamaK: Use either "in case" or "when", not both.

dbaeumer:
@doriath there is now a `PrepareRenameRequest` that can be sent from the client to the server to test if a position can be renamed. So I would like to only clarify the spec that `null` should be treated the same way as an empty edit. I would not bind this to newName is the same than oldName.

Can you adjust the PR accordingly ?

doriath:
dbaeumer@ I agree that there is now `PrepareRenameRequest`  that mostly solves this issue, but even if I control one side (client or server), the other side can still opt-out. That means that sometimes I will need to use just the `RenameRequest` and I want to make sure I can handle it properly on both sides.

Just so that I understand, is there another case where null / empty edit could be returned, other than newName equal oldName?

dbaeumer:
@doriath there is a difference between a rename request at an invalid position. In this case a server should return an error. If it did nothing (for whatever reason) then it should return null or an [] array.

dbaeumer:
Merged by hand into 3.17.

PR #637, opened 2018-12-12
Add optional reason for why the client failed to apply an edit

### LINKED ISSUE #618: ApplyWorkspaceEditResponse should contain a reason for why the edit was not applied
When sending an `applyEdit` request to the client, it can respond with a `ApplyWorkspaceEditResponse` that has a boolean field named `applied`:

```
	/**
	 * Indicates whether the edit was applied or not.
	 */
	applied: boolean;
```

There is no field for the client to supply a reason why the edit failed (for example, was it that the version of the document was wrong? did the edit reference invalid files? did it contain invalid positions?) which could make it difficult to track down when something goes wrong.

DanTup:
Fixes #618.

@dbaeumer I don't know if this is all that's required, or if my description is good enough - but feel free to suggest changes and/or tweak.

(It'd also be good for VS Code to use this too, for example describing a version mismatch if that's the reason it rejects edits).

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2417,6 +2417,14 @@ export interface ApplyWorkspaceEditResponse {
 	 * Indicates whether the edit was applied or not.
 	 */
 	applied: boolean;
+	
+	/**
+	 * A optional textual description for why the edit was not applied.
+	 * This may be used may be used by the server for diagnostic
+	 * logging or to provide a suitable error for a request that
+	 * triggered the edit.
+	 */
+	failureReason?: string;
 }
 ```
 * error: code and message set in case an exception happens during the request.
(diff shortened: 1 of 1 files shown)


dbaeumer:
@DanTup would be cool if you could add it to the protocol as well (it is here: https://github.com/Microsoft/vscode-languageserver-node/blob/master/protocol/src/protocol.ts#L1928)

DanTup:
@dbaeumer is this good? https://github.com/Microsoft/vscode-languageserver-node/pull/443

dbaeumer:
Yes. Thanks.

dbaeumer:
Sorry totally missed that.

Thanks for pinging.

PR #668, opened 2019-01-24
Rewording

king6cong:


dbaeumer:
I will close the PR since it is against an old unmaintained version of the spec.

PR #671, opened 2019-01-25
LSIF blog feedback

egamma:


egamma:
A first round of comments

mpickering:
The blog post explicitly mentions the use case of browsing code which hasn't been checked out but it isn't clear at all to me how to do this. Would it be possible to document it somewhere which can be linked from the post?

The conceptual problem I have is that it seems that you need to set the `uri` to something like `http://my-server.com/program.hs`. That part is fine, but then how to do you actually get vscode to make requests which mention this URL so that the information is served? Is it possible to open `http://my-server.com/program.hs` in vscode? I can't see how you can do that from the interface. 

(As a bonus, if there could be an example which combines a local checkout with peeking a definition in a remote file, that would help me understand a lot.)

dbaeumer:
@mpickering VS Code has support for file system providers we can provide file content for any form of URI's. You might want to have a look at https://marketplace.visualstudio.com/items?itemName=jrieken.vscode-memfs which is an extension implementing a FS provider. The actual full source is https://marketplace.visualstudio.com/items?itemName=jrieken.vscode-memfs

dbaeumer:
Closing PR. Changes got merged by hand.

PR #695, opened 2019-03-14
Fix number of references example in LSIF spec

arjun27:
This PR updates the number of references for `I#foo()` and `B#foo()` in the example elaborated for `referenceResults`


dbaeumer:
Thanks!

PR #701, opened 2019-03-20
Add missing declarationProvider field into server's capabilities

### LINKED ISSUE #673: textDocument/declaration missing a ServerCapabilities indicator
`definitionProvider`, `typeDefinitionProvider` etc exist - is the absence of `declarationProvider` an oversight?

Would be useful e.g. here: https://github.com/Valloric/ycmd/blob/5996d7daf2fda93ba2a0f6cfa86cf7c99e20170a/ycmd/completers/language_server/language_server_completer.py#L59

rcjsuen:
It seems like the `declarationProvider` field was forgotten when updating the protocol to 3.14.0. It exists in the [Microsoft/vscode-languageserver-node repository](https://github.com/Microsoft/vscode-languageserver-node/blob/dbfd37e35953ad0ee14c4eeced8cfbc41697b47e/protocol/src/protocol.declaration.ts#L38-L43) so I don't see why it shouldn't exist here.

This should close #673.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1866,6 +1866,12 @@ interface ServerCapabilities {
 	 */
 	foldingRangeProvider?: boolean | FoldingRangeProviderOptions | (FoldingRangeProviderOptions & TextDocumentRegistrationOptions & StaticRegistrationOptions);
 	/**
+	 * The server provides go to declaration support.
+	 *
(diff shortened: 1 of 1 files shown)


dbaeumer:
Thanks!

PR #709, opened 2019-03-29
Change character offsets from UTF-16 codeunits to Unicode codepoints

soc:
Addresses https://github.com/Microsoft/language-server-protocol/issues/376.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -227,7 +227,7 @@ type DocumentUri = string;
 
 #### Text Documents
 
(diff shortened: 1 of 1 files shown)


mikesamuel:
Should the commit message not note that this would break compatibility?
IIUC, *Position*s from clients written to version 3.16 would not work with servers complying with this change and vice versa.
The change in spec text should probably be accompanied by a non-normative note explaining how to manage that transition.  Without that, this would engender a lot of confusion.

dbaeumer:
@soc we can't simply change the spec that way since it is a breaking change and makes most existing server non spec conformant. The right approach is to add a capability on the client and server side and to spec a corresponding handshake.

soc:
@dbaeumer Did you check the date of this issue?

dbaeumer:
@soc Sorry, I reacted based on @mikesamuel comment. Any objection to close the PR since as noted this needs more work

mikesamuel:
Sorry for the confusion

axelson:
@dbaeumer I think closing this particular PR only makes sense if @soc doesn't plan to update the PR based on your comment (as well as fix the merge conflict). But would you consider a similar PR that does add a capability on the client and server side and specs a corresponding handshake?

soc:
Various implementations introduced some kind of encoding negotiation or changed the defaults already.

The world has moved on and the PR is obsolete.

dbaeumer:
Yes, I would accept a PR that defines such a handshake and adds an implementation to one of the client and server libraries that are out there.

PR #710, opened 2019-03-29
Update project vertex properties in LSIF spec

arjun27:
These changes to project vertex reflect the updated [protocol.ts](https://github.com/Microsoft/lsif-typescript/blob/master/shared/protocol.ts#L289).


dbaeumer:
@arjun27 Thanks!

PR #712, opened 2019-04-04
Update LSIF spec with tools

arjun27:
* Removes outdated protocol.ts file
* Links to new protocol package
* Links to updated `lsif-tsc` and `lsif-npm` tools


dbaeumer:
Thanks you very much.

PR #725, opened 2019-04-16
correct id in lsif textDocument/definition example

bmewburn:
Should this be 4? 14 is the resultSet -> definitionResult edge


dbaeumer:
Thanks!

PR #728, opened 2019-04-17
Clarify lazy results

LukaszMendakiewicz:
This applies what I believe was the intention of https://github.com/Microsoft/lsif-node/issues/3

Further changes in `protocol.ts` may be needed.


dbaeumer:
I updated `protocol.ts` were necessary (no property name required for result).

PR #735, opened 2019-04-28
Fix #375 Add more formatting options

### LINKED ISSUE #375: Add whitespace-related properties to FormattingOptions
```TypeScript
/**
 * Value-object describing what options formatting should use.
 */
interface FormattingOptions {
	/**
	 * Size of a tab in spaces.
	 */
	tabSize: number;

	/**
	 * Prefer spaces over tabs.
	 */
	insertSpaces: boolean;

	/**
	 * Signature for further properties.
	 */
	[key: string]: boolean | number | string;
}
```
Currently, `FormattingOptions` only defines two properties. I would like to propose that we add some additional properties (names can be reworded obviously) such as the following:

```TypeScript
interface FormattingOptions {
    /**
     * Trim trailing whitespaces on a line.
     */
    trimTrailingWhitespace: boolean;

    /**
     * Insert a newline character at the end of the file if one does not exist.
     */
    insertFinalNewline: boolean;

    /**
     * Trim all newlines after the final newline at the end of the file.
     */
    trimFinalNewlines: boolean;
}
```

rcjsuen:
Introduce some more formatting options related to:
- inserting a newline at the end of the file
- trimming whitespace at the end of a line
- trimming excessive newlines at the end of the file

This fixes #375.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -4105,6 +4105,21 @@ interface FormattingOptions {
 	insertSpaces: boolean;
 
 	/**
+	 * Trim trailing whitespaces on a line.
+	 */
+	trimTrailingWhitespace?: boolean;
+
+	/**
+	 * Insert a newline character at the end of the file if one does not exist.
+	 */
+	insertFinalNewline?: boolean;
+
+	/**
+	 * Trim all newlines after the final newline at the end of the file.
+	 */
+	trimFinalNewlines?: boolean;
+
+	/**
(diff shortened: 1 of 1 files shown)


rcjsuen:
I didn't update the change log as I presume we want to update it only when everything for 3.15.0 is good to go.

dbaeumer:
Merged by hand into dbaeumer/3.15.

PR #741, opened 2019-05-07
Correct my typo

ShalokShalom:
Now its correct :)

Format of the markdown was incorrect due to one ) instead of ]


dbaeumer:
Thanks!!

PR #756, opened 2019-05-23
Adds performance section to LSIF guide

arjun27:
The section defines benchmark for LSIF (equal to compilation time) and the approach taken by lsif-tsc to compute references efficiently


dbaeumer:
Thanks

PR #770, opened 2019-06-10
Remove "non-empty string" comment for workspace/symbol

DanTup:
VS Code sends empty strings to LSP servers here to get a full symbol list. Based on discussion at https://github.com/microsoft/vscode-languageserver-node/issues/458 I believe that the VS Code behaviour is desired and the spec is incorrect (the original intention was designed to avoid clients accidentally requesting all symbols before the user had typed, but it seems leaving that to the server is the best option).

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2449,7 +2449,8 @@ _Request_:
  */
 interface WorkspaceSymbolParams {
 	/**
-	 * A non-empty query string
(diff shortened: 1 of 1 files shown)


dbaeumer:
Merged into dbaeumer/3.15 branch

PR #804, opened 2019-07-26
Use proper enum where appropriate

### LINKED ISSUE #679: documentSymbol's SymbolInformation kind is not specific
While working on a client implementation, I noticed that the `SymbolInformation` kind property is just a number. It includes no indication of the meaning of the value. In practice, it appears that servers are using `SymbolKind`.

Is this something that should be clarified in the specification?

KamasamaK:
- `DiagnosticSeverity` was defined, but unused for `Diagnostic.severity`.
- `SymbolKind` was unused for `SymbolInformation.kind`
- Corrected some spelling issues.

Fixes #679


dbaeumer:
Actually Diagnostic is done as well.

PR #805, opened 2019-07-27
Added SignatureHelpContext

### LINKED ISSUE #561: Provide Previously Select Item for Signature Help
Current the spec specifies that the active signature index defaults to 0 if it is omitted. In a sense the implementation must provide this value or it will simply use 0. The problem lies that the language server doesn't know which item was user selected, as this information lies with the editor and isn't forwarded to the server.

```TypeScript
/**
* The active signature. If omitted or the value lies outside the
* range of `signatures` the value defaults to zero or is ignored if
* `signatures.length === 0`. Whenever possible implementors should
* make an active decision about the active signature and shouldn't
* rely on a default value.
* In future version of the protocol this property might become
* mandatory to better express this.
*/
activeSignature?: number
```
https://microsoft.github.io/language-server-protocol/specification#textDocument_signatureHelp

This means that even if a user explicitly selects an option, and the signature matches for the first set of arguments. Then the server has to choose between two equally acceptable signatures and has no information that the user has explicitly selected one of the two signatures.

Forwarding which signature was previously selected would help keep the user's selection intact in the case that a better match isn't available.

```C++
void foo(int one, int two, double three);
void foo(int another, int somethingElse, int again);


void main()
{
	foo(0, // cursor here
}
```
As an example, even if the next parameter is given, both signatures are suitable for what has currently been written. But the server still has to make a selection, or the active signature will default to 0. The desired behavior is to keep the user's current selection in this case.

### LINKED ISSUE #797: SignatureHelpContext support
Just as there is a `CompletionContext`, there should also be a `SignatureHelpContext`. The [VS Code API](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/47c71265ab38ee27fdc43e916189990996af1852/types/vscode/index.d.ts#L3199) has this already. I've copied the definition below.

```typescript
interface SignatureHelpContext {
	/**
	 * Action that caused signature help to be triggered.
	 */
	readonly triggerKind: SignatureHelpTriggerKind;

	/**
	 * Character that caused signature help to be triggered.
	 *
	 * This is `undefined` when signature help is not triggered by typing, such as when manually invoking
	 * signature help or when moving the cursor.
	 */
	readonly triggerCharacter?: string;

	/**
	 * `true` if signature help was already showing when it was triggered.
	 *
	 * Retriggers occur when the signature help is already active and can be caused by actions such as
	 * typing a trigger character, a cursor move, or document content changes.
	 */
	readonly isRetrigger: boolean;

	/**
	 * The currently active [`SignatureHelp`](#SignatureHelp).
	 *
	 * The `activeSignatureHelp` has its [`SignatureHelp.activeSignature`] field updated based on
	 * the user arrowing through available signatures.
	 */
	readonly activeSignatureHelp?: SignatureHelp;
}
```

Resolves #561

KamasamaK:
This adds `SignatureHelpContext` to the `textDocument/signatureHelp` request.

Fixes #561

Fixes  #797

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1136,7 +1136,7 @@ export interface TextDocumentClientCapabilities {
 		 * The client supports did save notifications.
 		 */
 		didSave?: boolean;
-	}
+	};
 
 	/**
 	 * Capabilities specific to the `textDocument/completion`
@@ -1165,7 +1165,7 @@ export interface TextDocumentClientCapabilities {
 			/**
 			 * The client supports commit characters on a completion item.
 			 */
-			commitCharactersSupport?: boolean
+			commitCharactersSupport?: boolean;
 
 			/**
 			 * The client supports the following content formats for the documentation
@@ -1182,7 +1182,7 @@ export interface TextDocumentClientCapabilities {
 			 * The client supports the preselect property on a completion item.
 			 */
 			preselectSupport?: boolean;
-		}
+		};
 
 		completionItemKind?: {
 			/**
@@ -1196,7 +1196,7 @@ export interface TextDocumentClientCapabilities {
 			 * the initial version of the protocol.
 			 */
 			valueSet?: CompletionItemKind[];
-		},
+		};
 
 		/**
 		 * The client supports to send additional context information for a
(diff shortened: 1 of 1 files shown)


KamasamaK:
Note that I changed `SignatureHelpTriggerKind.Invoke` from VS Code to `Invoked` so it is consistent with `CompletionTriggerKind`, which was also changed from VS Code.

Also, sorry I included some unrelated syntax fixes that I noticed while doing this.

dbaeumer:
Merged in 3.15 branch. https://github.com/microsoft/language-server-protocol/tree/dbaeumer/3.15

dbaeumer:
Add implementation to client as well.

PR #820, opened 2019-09-08
Fix typos in gh-pages

### LINKED ISSUE #813: Specification typos
In the current (`68775c83e00e02d9e2972858b74cfe5d3c779c82`) `specification.md`:

https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#shutdown "they have sent a shutdown requests" ==> "they have sent a shutdown request".

https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#implementation-considerations "if a client notices that a server exists unexpectedly" ==> "if a client notices that a server exits unexpectedly"

ebkalderon:
### Fixed

* Correct "shutdown requests" to "shutdown request."
* Correct "exists" to "exits."

Closes #813.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1930,7 +1930,7 @@ interface InitializedParams {
 
 #### <a href="#shutdown" name="shutdown" class="anchor">Shutdown Request (:leftwards_arrow_with_hook:)</a>
 
-The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit. Clients must not send any notifications other than `exit` or requests to a server to which they have sent a shutdown requests. If a server receives requests after a shutdown request those requests should be errored with `InvalidRequest`.
+The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit. Clients must not send any notifications other than `exit` or requests to a server to which they have sent a shutdown request. If a server receives requests after a shutdown request those requests should be errored with `InvalidRequest`.
 
 _Request_:
 * method: 'shutdown'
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -1930,7 +1930,7 @@ interface InitializedParams {
 
 #### <a href="#shutdown" name="shutdown" class="anchor">Shutdown Request (:leftwards_arrow_with_hook:)</a>
 
-The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit. Clients must not send any notifications other than `exit` or requests to a server to which they have sent a shutdown requests. If a server receives requests after a shutdown request those requests should be errored with `InvalidRequest`.
+The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit. Clients must not send any notifications other than `exit` or requests to a server to which they have sent a shutdown request. If a server receives requests after a shutdown request those requests should be errored with `InvalidRequest`.
```
SCKelemen:
```suggestion
The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit. Clients must not send any notifications other than `exit` requests to a server to which they have sent a shutdown request. If a server receives requests after a shutdown request those requests should error with `InvalidRequest`.
```

specification.md
```diff
@@ -4364,7 +4364,7 @@ Language servers usually run in a separate process and client communicate with t
...
 - if servers end up in an inconsistent state they should log this to the client using the `window/logMessage` request. If they can't recover from this the best they can do right now is to exit themselves. We are considering an [extension to the protocol](https://github.com/Microsoft/language-server-protocol/issues/646) that allows servers to request a restart on the client side.
-- if a client notices that a server exists unexpectedly it should try to restart the server. However clients should be careful to not restart a crashing server endlessly. VS Code for example doesn't restart a server if it crashes 5 times in the last 180 seconds.
+- if a client notices that a server exits unexpectedly it should try to restart the server. However clients should be careful to not restart a crashing server endlessly. VS Code for example doesn't restart a server if it crashes 5 times in the last 180 seconds.
```
SCKelemen:
```suggestion
- if a client notices that a server exits unexpectedly, it should try to restart the server. However clients should be careful not to restart a crashing server endlessly. VS Code, for example, doesn't restart a server which has crashed 5 times in the last 180 seconds.
```

specification.md
ebkalderon: Nice find! Thanks.

specification.md
ebkalderon: Thanks for the suggestion!

dbaeumer:
Thanks. Merged by hand due to the reorg of the specification.md.

PR #838, opened 2019-10-12
link specification.md to the current specification.

xuhdev:


dbaeumer:
@xuhdev can you please explain why who think this is necessary?

xuhdev:
One may want a link to the latest spec at all time, as a general information matter. For example, this page https://langserver.org/ (note there is a link to LSP spec)

dbaeumer:
I added this using a forward managed by `jekyll-redirect-from`

PR #839, opened 2019-10-15
Language Server references for COBOL and PL/I

phaumer:
Added entries for IBM's free COBOL and PL/I language servers.


phaumer:
The sorting is by language and not product name as people would look for the language.

PR #866, opened 2019-11-26
Correct misuse of "then" where "than" is appropriate.

calcnerd256:
The commit at the head of the gh-pages branch at the time of this writing says "different ... then" instead of "different ... than" inside of a conditional clause.  I have changed "then" to "than" and also made explicit the implicit "then" of the condition, though I am perfectly willing to allow the latter to remain implicit if people prefer that.


calcnerd256:
I forgot to open an issue before sending this pull request.  See #867.

PR #916, opened 2020-02-14
Replace ts with typescript in 3.15 markdown

reznikmm:
to be uniform with others typescript snippets


dbaeumer:
Nice cleanup. Thanks!

PR #930, opened 2020-02-19
Add workspaceFolders/configuration/window to 3.15

reznikmm:
It seems `workspaceFolders` and `configuration` properties were lost, but they are present in 3.14.
The `window` property is new in 3.15, let's add it to `ClientCapabilities` definition


dbaeumer:
Thanks again ! Merged by hand due to conflicts.

You might want to think about signing the CLA if you plan to provide more PRs.

PR #936, opened 2020-03-03
Add ocamllsp to list of lsp servers

rgrinberg:


_implementors/servers.md
```diff
@@ -74,6 +74,7 @@ index: 1
 | [MSBuild](https://github.com/Microsoft/msbuild) | [Adam Friedman](https://github.com/tintoy) | [msbuild-project-tools-vscode](https://github.com/tintoy/msbuild-project-tools-vscode) | TypeScript |
 | Nim | [Peter Munch-Ellingsen](https://github.com/PMunch) | [nimlsp](https://github.com/PMunch/nimlsp) | Nim |
 | OCaml/Reason | [freebroccolo](https://github.com/freebroccolo) | [ocaml-language-server](https://github.com/freebroccolo/ocaml-language-server) | TypeScript |
+| OCaml/Reason | [OCaml Labs](http://ocamllabs.io) | [ocamllsp](https://github.com/ocaml/ocaml-lsp) | OCaml |
```
rcjsuen: Should the entry above this be deleted? 🤔

rcjsuen: Actually, I guess they aren't the same project? I thought the older one got moved somewhere else (since the link no longer works) but I guess they're not the same thing as the old one was written in TypeScript and this other one is written in OCaml?

rgrinberg:
I would certainly recommend the lsp server that we maintain at OCaml Labs. However, it’s probably better if @freebroccolo decides on whether his server should be removed from the list.

dbaeumer:
I will merge the PR then.

_implementors/servers.md
KamasamaK: It looks like it was moved to https://github.com/ocaml-lsp/ocaml-language-server, and yes they're different projects.

Issue #943, opened 2020-03-09
Meaning of trace for the server


giltho:
The trace parameter is used for the clients (or at least VSCode) to write down the trace of execution. What is the point for the server ? Should it be the level of logging for `logMessage` ?
The only information the spec gives is:
```ts
/**
  * The initial trace setting. If omitted trace is disabled ('off').
  */
trace?: 'off' | 'messages' | 'verbose';
```
Given an explanation, I'd be glad to write a PR that gives more details about this parameter in the spec :)

dbaeumer:
Actually this deserves some clarifications. The LSP server I maintain send a custom `$/logTraceNotification` notification to the client. This works since the spec says 

> If a server or client receives notifications starting with ‘$/’ it is free to ignore the notification

However I think it makes sense to standardize at least the method and the payload. Are you willing to look into this?

giltho:
I would be glad to.
However, I would need a bit more information.

I'm helping in the [ocaml lsp implementation](https://github.com/ocaml/ocaml-lsp), and we write a lot of logs through [window/logMessage](https://microsoft.github.io/language-server-protocol/specification#window_logMessage) since [this recent PR](https://github.com/ocaml/ocaml-lsp/pull/120). We were considering logging even more things through this process. However, I kind of wondered:
 - Should part of what we report be reported through a trace thing ?
 - Should the `trace` setting impact how much we log ?

In a more general way, I like the idea of standardising your `$/logTraceNotification` notif, but I have no idea what kind of payload it should carry. Do you have a link to your implementation so I could understand better ?

dbaeumer:
Regarding your questions:

> Should part of what we report be reported through a trace thing ?
Yes, I think so.
> Should the trace setting impact how much we log ?
Yes, since it support verbose, messages, off

The notification is typed here: https://github.com/microsoft/vscode-languageserver-node/blob/dbaeumer/encodingSupport/jsonrpc/src/main.ts#L271

`message` is the message
`verbose` and extract verbose information

giltho:
Sorry to be a bit slow to answer
So I think I could write this, I have one last question : when should `logMessage` be used, and when should `logTrace` be used ?

dbaeumer:
`logMessage` is more like `console.log`. `logTrace` is for writing a trace execution. Clients could for example write this using speedy log for example.

dbaeumer:
PR got merged.

PR #953, opened 2020-04-05
Specify logTrace notification 

giltho:
Hi ! This is a follow-up on #943 

This is just a draft since :
- [x] I'd like to add a sentence about the difference between logTrace and logMessage
- [x] Reference implementation
- [x] I do now know if `logTrace` should be a "general" notification or under the `window` scope
- [x] There should probably be a client capability added for retro-compatibility. In the case of ocaml-lsp, I'd like to know, depending on the client, if I continue to send `logMessage` notifications, or if I can send `logTrace` notifications instead and they will be supported.

### DIFF
--- a/_specifications/specification-3-15.md
+++ b/_specifications/specification-3-15.md
@@ -1289,15 +1289,6 @@ export interface PartialResultParams {
 }
 ```
 
-#### <a href="#traceValue" name="traceValue" class="anchor"> TraceValue </a>
-
-A `TraceValue` represents the level of verbosity with which the server systematically reports its execution trace using [$/logTrace](#logTrace) notifications.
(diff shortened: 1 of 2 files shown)


giltho:
Added a paragraph that also specifies setTrace, but it can be easily reverted if this is not desired

giltho:
Hi @dbaeumer, is there anything else you'd like added to this PR ? :)
This would be a great feature to have officially in `ocaml-lsp`

dbaeumer:
@giltho sorry for that. Only one thing: we can't change the existing 3.15 spec. The changes you dod should go into the 3-16 version here: https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#L9

giltho:
@dbaeumer Nothing to be sorry about 😄 
I've moved the changes to 3.16, so it should be ok now

PR #955, opened 2020-04-07
Add lsp-types rust implementation

kjeremy:


mkurnikov:
Maybe add https://github.com/rust-analyzer/lsp-server too?

kjeremy:
@mkurnikov I thought about it but don't think that it really fits as an "sdk".

PR #1000, opened 2020-05-24
Use sticky table headers

### LINKED ISSUE #985: [website] use sticky table headers
#### The problem
When I visit https://microsoft.github.io/language-server-protocol/implementors/servers/ to check the implementation table, I often have to scroll back up to the top of the table to check the headers.  If I'm checking an implementation low on the table, I'll see something like 

```
| ... 
+------------+--------+-------+------------+
| a language | a name | a url | a language |
+------------+--------+-------+------------+
| ...
```

which hides which column represents what.  Since there's no natural order to the list of features implemented, I'm unlikely to remember the column order.
#### Proposed solution
Use a `position: sticky` header on the `<thead>`s of the large tables.

Similar to https://github.com/langserver/langserver.github.io/issues/160.

SKalt:
Closes #985. I'm pleased to note that sticky table headers are useful in the specification and SDKs page as well. 

Notes on the implementation: the `table.table-responsive { display: block }` from bootstrap prevented sticky table headers from sticking.  I've checked around and I can't notice any adverse effects from moving back to displaying tables as tables, but I'd recommend reviewers check for unexpected side-effects as well.

### DIFF
--- a/css/main.scss
+++ b/css/main.scss
@@ -54,7 +54,8 @@ table.table-responsive.table-bordered {
 			position: -moz-sticky;
 			position: -ms-sticky;
 			position: -o-sticky;
-			top: 55px;
+			top: 3rem; // 2.5rem for the header's height + padding,
+			           // +.5rem for the tr's own padding 
 			background: #fff;
 			outline: 1px solid #e9ecef;
 		}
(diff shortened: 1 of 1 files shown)


dbaeumer:
@SKalt THANKS!

PR #1012, opened 2020-06-06
ResponseMessage.result can be an array type

yxrkt:
Certain responses state that they can return arrays (Document Color Request, for example).

### DIFF
--- a/_specifications/specification-3-15.md
+++ b/_specifications/specification-3-15.md
@@ -110,7 +110,7 @@ interface ResponseMessage extends Message {
 	 * The result of a request. This member is REQUIRED on success.
 	 * This member MUST NOT exist if there was an error invoking the method.
 	 */
-	result?: string | number | boolean | object | null;
+	result?: string | number | boolean | array | object | null;
 
 	/**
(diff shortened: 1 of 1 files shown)


dbaeumer:
Correct. Merged by hand into 3.17.

darichey:
@dbaeumer Was this merged correctly? The current spec doesn't reflect the change.

KamasamaK:
The commit history does not show that this was ever corrected for 3.17.

dbaeumer:
Actually my conclusion was that `object` in TS includes `array` and therefore didn't add it. But this might not be obvious for everybody. But adding it might include confusion as well. May be we should add a comment somewhere.

darichey:
That's fair, but `array | object` appears in other places in the spec (`RequestMessage.params`, `ResponseError.data`, `NotificationMessage.params`). 

Personally, I think that they should be consistent and that explicitly including `array` is more clear.

dbaeumer:
Good point.

dbaeumer:
Added `array`

PR #1019, opened 2020-06-13
Document de-facto usage of lowercased driver letters

matklad:
Rust canonicalizes drive letters to the upper case, and that caused a hard-to-debug bug. In the ideal world, both servers and clients would do the actual canonicalization themselves, but I think it's a good idea to document status quo.

### DIFF
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -249,6 +249,8 @@ scheme     authority       path        query   fragment
 
 We also maintain a node module to parse a string into `scheme`, `authority`, `path`, `query`, and `fragment` URI components. The GitHub repository is [https://github.com/Microsoft/vscode-uri](https://github.com/Microsoft/vscode-uri) the npm module is [https://www.npmjs.com/package/vscode-uri](https://www.npmjs.com/package/vscode-uri).
 
+On case-insensitive file systems, two textually different URIs might represent the same file.
+
 Many of the interfaces contain fields that correspond to the URI of a document. For clarity, the type of such a field is declared as a `DocumentUri`. Over the wire, it will still be transferred as a string, but this guarantees that the contents of that string can be parsed as a valid URI.
 
 ```typescript


matklad:
Hm, I wonder if VS Code side was actually fixed by https://code.visualstudio.com/updates/v1_46#_file-path-casing?

dbaeumer:
Actually https://code.visualstudio.com/updates/v1_46#_file-path-casing solves the problem of having two URLs with different casing pointing to the same file in case insensitive file systems. 

But the overall problem still exists and it is actually not bound to the drive letter. I think instead of saying that the drive letter should be lower cases we should say something about the fact that a server can't assume that a file URI represents the file on disk in case of casing. So they need to do their own normalization if they for example use these URIs as keys in hashtables.

matklad:
>that a server can't assume that a file URI represents the file on disk in case of casing

You mean, *both* the server and the client? IIRC, our original problem was that VS Code didn't understand the URI we were sending back.

dbaeumer:
On systems that are case insensitive each component needs to assume that the file URI is represented in a different case then the one you already have for the file. Otherwise we would need to provide a central service to normalize them. 

What changed in VS Code is that fact that if VS Code has a file under the URI `file:///abc/text.txt` and you now send it as `file:///aBc/TeXt.txt` then VS Code will detect that it is the same file on Windows and for example if you generate a diagnostic for it and click on it it will open the correct file only once.

matklad:
Changed to a general note about being aware about the casing. I thought of requiring both the server and the client to be completely case-agnostic for case-insensitive systems, but that is awkward:

* I think figuring out if a system is case-sensitive is actually pretty complicated (I think one can mount both case-sensitive and case-insensitive file systems at the same time, also the case tables themselves depend on a specific file systems).
* I think in some cases it might makes sense to make some assumptions about the case? For example, I think it's reasonable to assume that `DidOpenTextDocument` and subsequent `DidChangeTextDocument` use the same URI, so that using a hash-map keyed by URI for the set of opened files should be allowed.

_specifications/specification-3-16.md
```diff
@@ -249,6 +249,8 @@ scheme     authority       path        query   fragment
 
 We also maintain a node module to parse a string into `scheme`, `authority`, `path`, `query`, and `fragment` URI components. The GitHub repository is [https://github.com/Microsoft/vscode-uri](https://github.com/Microsoft/vscode-uri) the npm module is [https://www.npmjs.com/package/vscode-uri](https://www.npmjs.com/package/vscode-uri).
 
+On case-insensitive file systems, two textually different URIs might represent the same file.
```
mikesamuel, line 252: Could this be rephrased to avoid leading someone to conclude that on case-sensitive file systems, textually different URIs represent different files?

These URIs are textually different but [*equivalent*](https://url.spec.whatwg.org/#url-equivalence) despite using only one case:
- `file:///foo/bar`
- `file:///foo/./bar`
- `file:///foo/%62ar`
- `file://localhost/foo/bar

Even ignoring equivalence, is this not also true of all file systems that allow symlinks or multiple hard links?

matklad: I guess at this point it stalled long enough to just close.

PR #1068, opened 2020-08-12
Add capabilities fields for semantic tokens

DanTup:
These fields were documented, but not included in this definition. This prevented them showing up in the code I'm generating from the spec.


dbaeumer:
Thanks. Forgot about them.

Issue #984, opened 2020-05-13
Support onWillRenameFiles to allow automatic updates to imports


DanTup:
Similar to #873, but for *will* instead of *did*. This is the equiv of `onWillRenameFiles: Event<FileWillRenameEvent>` in VS Code.

Some language servers (including Dart) need to build the edits for updating import statements *before* the files are moved, so only telling the server after the event would not be enough.

dbaeumer:
Anyone will to work on a PR on this?

ericdallo:
That is a feature that we really miss on [lsp-dart](https://emacs-lsp.github.io/lsp-dart/), it would be a great addition to the spec.

DanTup:
@dbaeumer would something like this work? In VS Code the event has a `waitUntil` method which I don't think LSP can represent, so the options are probably limiting this to returning a workspace edit, or allowing the server to do its own work (eg. sending `workspace/applyEdit`s) and the client waiting on a response before performing the renames.

## WillRenameFiles Request

The `workspace/willRenameFiles` request is sent from the client to the server to inform the server about file/folder renames that are about to occur. The client will not perform the rename until the server has responded to this request and the client has processed any server-to-client requests sent prior to the response.

This request supports both files and folders therefore it is the servers responsibility to determine if any work needs to be done for a given rename (for example, whether a folder being renamed contains any files of interest).

A possible flow for this request would be:

- User invokes a file/folder rename action in the client
- Client sends a `workspace/willRenameFiles` request to the server but does not rename any files
- Server computes edits to update references to the moved files
- Server sends edits to the client using `workspace/applyEdit`
- Server waits for client to acknowledge the edits were applied
- Server responses to the `workspace/willRenameFiles` request
- Upon receiving the response, the client renames the folder/files

*Client Capability*:

- property (optional): `workspace.willRenameFiles`
- property type: `boolean`

The capability indicates that the client can send `workspace/willRenameFiles` requests.

*Server Capability*:

- property (optional): `workspace.willRenameFiles`
- property type: `boolean`

The capability indicates that the server can handle `workspace/willRenameFiles` requests.

**Request**:

- method: `workspace/willRenameFiles`
- params: `WillRenameFilesParams` defined as follows:

```ts
export interface WillRenameFilesParams {
	event: FileWillRenameEvent;
}

export interface FileWillRenameEvent {
	files: FileRename[];
}

export interface FileRename {
	oldUri: string;
	newUri: string;
}
```

**Response**:

[...]

PR #1074, opened 2020-08-24
Add window/{did,will}{Create,Rename,Delete}Files

DanTup:
@dbaeumer this is a first go at  #984 using `WaitUntil` requests. It currently only includes rename, but we can duplicate for create/delete once these are good if required.

Some things I wasn't sure about:

- Should the capabilities just be bools, or additional nested objects for easier extending in future (or would they just be changed to `boolean | {/*...*/}`?
- Is `workspace` the right place for these? Save etc. are in `textDocument/` but these could be folder renames so there's no specific document.
- Is it valid to have no registration options? Servers should be able to opt in/out, but otherwise I don't think there's anything (I don't think using document selectors would make sense as it may force the client to scan down the whole tree for folder renames).
- Is v3.16.0 correct or may that release before this is done?

(I haven't tried to generate my code from these changes yet, but I'd like to do so before this is merged - but I figured there may be some iteration first)

### DIFF
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -3222,7 +3222,7 @@ _Client Capability_:
 * property name (optional): `window.fileOperations.willCreate`
 * property type: `boolean`
 
-The capability indicates that the client supports `window/willCreateFiles` requests.
+The capability indicates that the client supports sending `window/willCreateFiles` requests.
 
 _Server Capability_:
 * property name (optional): `window.fileOperations.willCreate`
@@ -3244,7 +3244,7 @@ interface FileOperationRegistrationOptions {
 }
 ```
 
-The capability indicates that the server is interested in `window/willCreateFiles` requests.
+The capability indicates that the server is interested in receiving `window/willCreateFiles` requests.
 
 _Registration Options_: none
 
@@ -3285,13 +3285,13 @@ _Client Capability_:
 * property name (optional): `window.fileOperations.didCreate`
 * property type: `boolean`
 
-The capability indicates that the client supports `window/didCreateFiles` notifications.
+The capability indicates that the client supports sending `window/didCreateFiles` notifications.
 
 _Server Capability_:
 * property name (optional): `window.fileOperations.didCreate`
 * property type: `FileOperationRegistrationOptions`
 
-The capability indicates that the server is interested in CreateFiles` notifications.
+The capability indicates that the server is interested in receiving `window/didCreateFiles` notifications.
 
 _Notification_:
 * method: 'window/didCreateFiles'
@@ -3305,13 +3305,13 @@ _Client Capability_:
 * property name (optional): `window.fileOperations.willRename`
 * property type: `boolean`
 
-The capability indicates that the client supports `window/willRenameFiles` requests.
+The capability indicates that the client supports sending `window/willRenameFiles` requests.
 
 _Server Capability_:
 * property name (optional): `window.fileOperations.willRename`
 * property type: `FileOperationRegistrationOptions`
 
-The capability indicates that the server is interested in `window/willRenameFiles` requests.
+The capability indicates that the server is interested in receiving `window/willRenameFiles` requests.
 
 _Registration Options_: none
 
@@ -3357,13 +3357,13 @@ _Client Capability_:
 * property name (optional): `window.fileOperations.didRename`
 * property type: `boolean`
 
-The capability indicates that the client supports `window/didRenameFiles` notifications.
+The capability indicates that the client supports sending `window/didRenameFiles` notifications.
 
 _Server Capability_:
 * property name (optional): `window.fileOperations.didRename`
 * property type: `FileOperationRegistrationOptions`
 
-The capability indicates that the server is interested in `window/didRenameFiles` notifications.
+The capability indicates that the server is interested in receiving `window/didRenameFiles` notifications.
 
 _Notification_:
 * method: 'window/didRenameFiles'
@@ -3377,13 +3377,13 @@ _Client Capability_:
 * property name (optional): `window.fileOperations.willDelete`
 * property type: `boolean`
 
-The capability indicates that the client supports `window/willDeleteFiles` requests.
+The capability indicates that the client supports sending `window/willDeleteFiles` requests.
 
 _Server Capability_:
 * property name (optional): `window.fileOperations.willDelete`
 * property type: `FileOperationRegistrationOptions`
 
-The capability indicates that the server is interested in `window/willDeleteFiles` requests.
+The capability indicates that the server is interested in receiving `window/willDeleteFiles` requests.
 
 _Registration Options_: none
 
@@ -3424,13 +3424,13 @@ _Client Capability_:
 * property name (optional): `window.fileOperations.didDelete`
 * property type: `boolean`
 
-The capability indicates that the client supports `window/didDeleteFiles` notifications.
+The capability indicates that the client supports sending `window/didDeleteFiles` notifications.
 
 _Server Capability_:
 * property name (optional): `window.fileOperations.didDelete`
 * property type: `FileOperationRegistrationOptions`
 
-The capability indicates that the server is interested in `window/didDeleteFiles` notifications.
+The capability indicates that the server is interested in receiving `window/didDeleteFiles` notifications.
 
 _Notification_:
 * method: 'window/didDeleteFiles'
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -7252,9 +7252,9 @@ Language servers usually run in a separate process and client communicate with t
 * Add support for code action resolve request.
 * Add support for diagnostic `data` property.
 * Add support for signature information `activeParameter` property.
-* Add support for `workspace/onDidCreateFiles`, `workspace/onWillCreateFiles`
-* Add support for `workspace/onDidRenameFiles`, `workspace/onWillRenameFiles`
-* Add support for `workspace/onDidDeleteFiles`, `workspace/onWillDeleteFiles`
+* Add support for `workspace/didCreateFiles` notifications and `workspace/willCreateFiles` requests.
+* Add support for `workspace/didRenameFiles` notifications and `workspace/willRenameFiles` requests.
+* Add support for `workspace/didDeleteFiles` notifications and `workspace/willDeleteFiles` requests.
 
 #### <a href="#version_3_15_0" name="version_3_15_0" class="anchor">3.15.0 (01/14/2020)</a>
 
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -3375,8 +3375,22 @@ interface FileOperationPattern {
 	 *
 	 * Matches both if undefined.
 	 */
-	matches?: 'file' | 'folder';
+	matches?: FileOperationPatternKind;
 }
+
+export namespace FileOperationPatternKind {
+	/**
+	 * The pattern matches a file only.
+	 */
+	export const file: 'file' = 'file';
+
+	/**
+	 * The pattern matches a folder only.
+	 */
+	export const folder: 'folder' = 'folder';
+}
+
+export type FileOperationPatternKind = 'file' | 'folder';
 ```
 
 The capability indicates that the server is interested in receiving `workspace/willCreateFiles` requests.
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -2879,13 +2879,13 @@ export interface ApplyWorkspaceEditResponse {
 The will rename files notification is sent from the client to the server before files are actually renamed.
 
 _Client Capability_:
-* property name (optional): `workspace.renameFiles.willRenameFiles`
+* property name (optional): `workspace.files.willRename`
 * property type: `boolean`
 
 The capability indicates that the client supports `workspace/willRenameFiles` notifications.
 
 _Server Capability_:
-* property name (optional): `workspace.renameFiles.willRenameFiles`
+* property name (optional): `workspace.files.willRename`
 * property type: `bool`
 
 The capability indicates that the server is interested in `workspace/willRenameFiles` notifications.
@@ -2927,13 +2927,13 @@ export namespace FileRename {
 The will rename files wait until request is sent from the client to the server before files are actually renamed. The request can return a WorkspaceEdit which will be applied to workspace before the files are renamed. Please note that clients might drop results if computing the edit took too long or if a server constantly fails on this request. This is done to keep renames fast and reliable.
 
 _Client Capability_:
-* property name (optional): `workspace.renameFiles.willRenameFilesWaitUntil`
+* property name (optional): `workspace.files.willRenameWaitUntil`
 * property type: `boolean`
 
 The capability indicates that the client supports `workspace/willRenameFilesWaitUntil` requests.
 
 _Server Capability_:
-* property name (optional): `workspace.renameFiles.willRenameFilesWaitUntil`
+* property name (optional): `workspace.files.willRenameWaitUntil`
 * property type: `boolean`
 
 The capability indicates that the server is interested in `workspace/willRenameFilesWaitUntil` requests.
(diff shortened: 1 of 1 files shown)


_specifications/specification-3-16.md
```diff
@@ -1591,6 +1591,28 @@ interface ClientCapabilities {
 		*/
 		workspaceFolders?: boolean;
 
+		/**
+		* The client has support for rename requests/notifications.
+		*
+		* Since 3.7.0
```
ericdallo, line 1597: Maybe this should be 3.17, right? 
Latest spec release is [3.16](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/)

DanTup: Oops, not sure where I got this. I think it should actually be v3.16 as that's still unreleased. I've updated it (and added to changelog). Thanks!

dbaeumer:
Have not forgotten you, however I need to push this to the beginning of next week. We have endgame.

DanTup:
@dbaeumer understood, np!

_specifications/specification-3-16.md
```diff
@@ -1591,6 +1591,28 @@ interface ClientCapabilities {
...
 		workspaceFolders?: boolean;
 
+		/**
+		* The client has support for rename requests/notifications.
+		*
+		* Since 3.16.0
+		*/
+		renameFiles?: {
+			/**
+			 * The client has support for didRenameFiles notifications.
```
kjeremy, line 1601: Maybe something like: "The client supports sending didRenameFiles notifications." so it's obvious which direction the communication flows?

DanTup:
@dbaeumer I don't suppose you've had chance to look at this yet?

_specifications/specification-3-16.md
dbaeumer: Assuming that we will add support for create and delete as well (@DanTup would you be willing to do so) to make this a complete story we can either do

```typescript
files: {
   willCreate: boolean;
   willCreateWaitUntil: boolean;
   didCreate: boolean;
   willRename: boolean;
   willRename: boolean;
   willRenameWaitUntil: boolean;
   willDelete: boolean;
   willDeleteWaitUntil: boolean;
   didDelete: boolean;
}
```

or we could do 

```typescript
files: {
    create: {
        ...
    },
    rename: {
        ....
    },
    delete: {
    }
}
```

I am actually for the first option.

_specifications/specification-3-16.md
```diff
@@ -1835,6 +1857,28 @@ interface ServerCapabilities {
 		 * @since 3.6.0
 		 */
 		workspaceFolders?: WorkspaceFoldersServerCapabilities;
+
+		/**
+		 * The server is interested in rename notifications/requests.
+		 *
+		 * @since 3.16.0
+		 */
+		renameFiles?: {
```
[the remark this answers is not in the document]
dbaeumer: and same structure of the literal.

dbaeumer (changes requested):

dbaeumer:
@DanTup actually was on my list for this week. Added my comments.

DanTup:
Thanks!

I've done the renames and added create/delete, though I have a few questions:

> willRename, didRename

Is there actually any difference between these? Since it's all async, it seems like `willRename` is going to arrive at the server after the rename has probably occurred (unless using `willRenameWaitUntil`), so in what case would it be useful for a server to user `willRename` rather than the other two? Should each operation just be a `willXXXWaitUntil` and a `didXXX`?

> To be consistent we should also consider adding a dynamicRegistration property. But one for all three is enough. It is the same as for text document synchronization.

The text doc registrations are scoped to text documents (eg. by language ID) but it wasn't clear how this would work here. For example if I say I want renames for just a Dart file, what happens when a I rename a folder? Should it always be sent, or only if it has child Dart files?

dbaeumer:
That is a good point. It is very unlikely that someone is interested in `willRename` and `willRenameWaitUntil`. The problem is that `willRename` can be implemented as a notification where as `willRenameWaitUntil` must be implemented as a request/respone.

What we could do have one property `willRename` that takes the following values: `off`, `event`, `waitUntil`. The opern question would still be if we implement `event` as a notification or as a request/response as well. For simplicity I would opt for a request/response as well but would document that in the event case the client doesn't need wait for the response since it will be empty.

DanTup:
> That is a good point. It is very unlikely that someone is interested in `willRename` and `willRenameWaitUntil`. The problem is that `willRename` can be implemented as a notification where as `willRenameWaitUntil` must be implemented as a request/respone.

Right - but what advantage is there having `onWillRename` versus `onDidRename`? Both are notifications, and by the time the server receives `onWillRename` it must assume the rename may have already happened, and therefore it seems like it's identical in function to `onDidRename`?

I'm thinking that we only need `onDidRename` and `onWillRenameWaitUntil`?

dbaeumer:
Regarding

>  For example if I say I want renames for just a Dart file, what happens when a I rename a folder? Should it always be sent, or only if it has child Dart files?

I think that should be consistent with file events. There a rename of a folder doesn't generate renames for the child files.

DanTup:
> That is a good point. It is very unlikely that someone is interested in `willRename` and `willRenameWaitUntil`. The problem is that `willRename` can be implemented as a notification where as `willRenameWaitUntil` must be implemented as a request/respone.

Right - but what advantage is there having `onWillRename` versus `onDidRename`? Both are notifications, and by the time the server receives `onWillRename` it must assume the rename may have already happened, and therefore it seems like it's identical in function to `onDidRename`?

I'm thinking that we only need `onDidRename` and `onWillRenameWaitUntil`?

> I think that should be consistent with file events. There a rename of a folder doesn't generate renames for the child files.

But would it be sent to the server if the server only asked for events for Dart files? (I think it has to - so I'm wondering if we need to specify that folder events will always be sent, regardless of the file/pattern filters on the registration? or there needs to be some way to register for them including folders?)

dbaeumer:
Regarding dynamic registration: I think there is a misunderstanding. Dynamic registration means that a server later on can register for these events and not only through the initialize result. It doesn't mean that it needs a document filter for registration. For these event I would argue that we either 
- only provide a file pattern to make it like the file events. 
- make the global.

These events WILL ONLY be fired if a new file is created, rename, deleted using a user action in the UI. They are not a replacement for normal file events.

Regarding which events we need: I agree that a `willRename` followed by a `didRename` doesn't make much sense. However I can easily argue that a `willRename` alone could make sense.

If we make the registration file pattern sensitive (please note that the VS Code API is not that way so we need to implement this ourselves) then I would do the following:

- a `will*` request/response
- a `did*` notification

If we make it global I would argue that we should not make all `will` a request/response since it might result in a performance problem if many server have registered for these events.

I am actually leaning towards the registration on a file pattern and then make the `will*` always a request/response. The name could simply be `willRename`.

DanTup:
> It doesn't mean that it needs a document filter for registration

Sorry - I misunderstood your comment about copying text document synchronisation.

> - only provide a file pattern to make it like the file events.
> - make the global

I think the first could be a little confusing - if I say I want just "*.dart" and the user renames a folder with a dart file nested inside it, it's not clear whether I should get the notification or not?

Also - would the pattern be applied to the old URI or new URI in the case of renames?

> If we make the registration file pattern sensitive

Should this say case-sensitive? (this might be a can of worms 😬)

> make the `will*` always a request/response. The name could simply be `willRename`.

So if I understand correctly, we will drop the current `willRename` (since it's basically the same as `didRename` and rename the `willRenameWaitUntil` to `willRename` in its place?

So we're left with:

- didRename (notification)
- willRename (request/response that waits)

(and the same for create/delete)?

dbaeumer:
I think this is not that easy because of

> If we make it global I would argue that we should not make all will a request/response since it might result in a performance problem if many server have registered for these events.

If we make `willRename` a request/response that waits and we have 5 servers registered globally for that event VS Code needs to wait for all 5 servers to respond before the actual rename can happen. And it is very likely that at the most one server does something.

In the VS Code API this is different since VS Code talks to one extension host and then calls in memory 5 listeners. If we go for a registration with a pattern we could clearly spec that:

- if a folder renames no events are generated for the files inside the folders.
- for rename the pattern fires for the oldURI.

Unclear is whether we need to allow to specify a pattern denoting a directory.

DanTup:
> If we make `willRename` a request/response that waits and we have 5 servers registered globally for that event VS Code needs to wait for all 5 servers to respond before the actual rename can happen. And it is very likely that at the most one server does something.

Fair point. Though I think most language servers would use `didRename`, since many implemented updating references before `willRename` was added to VS Code (I opened the original request, and don't think many other servers needed to know about the rename before it happened). I do agree that having to stall the client waiting for a response makes it less desirable to send events that won't do anything though.

> Unclear is whether we need to allow to specify a pattern denoting a directory.

I'm not sure if this means "do we need directory support" or "does it need to be a pattern"? I think directories could *probably* be just a flag (I suspect most servers will just want them all), but in the interest of allowing it to be controlled (for the same perf reasons mentioned above), maybe it should be two patterns (a file pattern, and a directory pattern)? For Dart, I think what I'm interested in is

files: **/*.dart - any Dart file rename
directories: **/* - all directories

How about something like this?

```ts
files?: {
	// Req/Resp returning TextEdit[] | null
	willCreate?: FileOperationRegistrationOptions;
	// Notification
	didCreate?: FileOperationRegistrationOptions;

	// Req/Resp returning TextEdit[] | null
	willRename?: FileOperationRegistrationOptions;
	// Notification
	didRename?: FileOperationRegistrationOptions;

	// Req/Resp returning TextEdit[] | null
	willDelete?: FileOperationRegistrationOptions;
	// Notification
	didDelete?: FileOperationRegistrationOptions;
}

interface FileOperationRegistrationOptions {
	documentSelector?: DocumentSelector;
	directorySelector?: DirectorySelector;
}

export type DirectorySelector = DirectoryFilter[];

export interface DirectoryFilter {
	pattern?: string;
}
```

DanTup:
@dbaeumer any thoughts on the comment above?

dbaeumer:
@DanTup was on vacation. Will have a look today.

DanTup:
@dbaeumer np! Wasn't sure if the notification may have been overlooked. Thanks!

dbaeumer:
@DanTup I do like the API. However IMO we should not talk about document selectors / filters. At least in VS Code to correctly determine a document selector you need to `open` the document to determine its language ID. I would argue we should treat these like file events which simply support a glob pattern. See https://github.com/microsoft/vscode-languageserver-node/blob/master/protocol/src/common/protocol.ts#L1419

Then we could say a pattern with a slash at the end matches a directory only.

DanTup:
@dbaeumer got it - I'll use `globPattern` then.

> Then we could say a pattern with a slash at the end matches a directory only.

Just to make sure I understand correctly, to support "all .dart files and any folder" I would therefore send two globs, like:

- **/*.dart
- **/

Is that right? If so, I'll update the Pr to match my previous comment above with this change for a proper review. Thanks!

dbaeumer:
@DanTup you can use two registrations or some OR pattern. Something like `**{/,*.dart}`

DanTup:
Ah cool, even better. I'll get this updated today. Thanks!

DanTup:
@dbaeumer I think this is good for a proper look now. Thanks!

dbaeumer:
@DanTup this looks OK to me. Are you going to work on an implementation PR in `vscode-languageserver-node` as well.

DanTup:
@dbaeumer yep, I'll have a go. I might shout if (when) I get stuck (though I think it should be mostly the same as `willSaveWaitUntil`).

DanTup:
@dbaeumer are there any notes on how you usually work on changes that span protocol/client? It looks like I need to put the type changes into `./protocol`, but the client library seems to be using the one from NPM. Is there an easy way to have it use the local folder while working on it? (this is probably more of an NPM question... in Dart there are dependency_overrides for this).

DanTup:
@dbaeumer I've made a start at trying to implement one of the 6 requests (`willCreateFiles`) here:

https://github.com/DanTup/vscode-languageserver-node/commit/498507ab4ef969f68990492480f7dad8f32fdf3a

Hopefully I'm going in the right direction, but I have some questions:

1. When I save server/client.ts, all of the formatting gets badly messed up (for example tons of blank lines written between all of the imports). I've tried disabling `formatOnSave`/`formatOnPaste`/`formatOnType` and the `codeActionsOnSave` in the Workspace settings, and yet it continues. I have to use "Save without formatting" every time and it's a real pain. Am I missing something? :(

2. In the case where a server registers statically, is just using a glob of `**/*` reasonable? (https://github.com/DanTup/vscode-languageserver-node/commit/498507ab4ef969f68990492480f7dad8f32fdf3a#diff-1442f78812d67be25f18d3b5d82e8b4191ae2f939a900eac32d4360222c5d845R1239)

3. Is there any existing code I can use for the glob matching? It seems like the existing code for the text document events uses VS Code's `Language.match` but it doesn't seem like the glob matching is exposed directly. Should I add the `glob` NPM package and use it directly? (https://github.com/DanTup/vscode-languageserver-node/commit/498507ab4ef969f68990492480f7dad8f32fdf3a#diff-1442f78812d67be25f18d3b5d82e8b4191ae2f939a900eac32d4360222c5d845R1257)

4. Should middleware use VS Code event types or LSP events? (https://github.com/DanTup/vscode-languageserver-node/commit/498507ab4ef969f68990492480f7dad8f32fdf3a#diff-1442f78812d67be25f18d3b5d82e8b4191ae2f939a900eac32d4360222c5d845R1267)

DanTup:
@dbaeumer I know you're busy (and I'm not helping that 😬), but interested in your thoughts on the above - would love to get more done on this. Thanks! :-)

dbaeumer:
Really busy week :-). Regarding your questions:

1.) the only thing that is enabled for the Node repository is code action on save. Could it be that you have an extension installed (e.g. prettier) that does this and is not happy with how the Node repository is formatted.
2.) This sounds reasonable to me.
3.) You can copy the code here: https://github.com/microsoft/vscode-eslint/blob/master/client/src/utils.ts#L184 I have written that a while back ago.
4.) The middle ware uses VS Code types for requests going from the client to the server and LSP types for things flowing from the server to the client. This allows incepting the data and modifying it before it is processed.

DanTup:
@dbaeumer as far as I can tell, that pattern matcher doesn't support slashes or stars inside braces:

https://github.com/microsoft/vscode-eslint/blob/master/client/src/utils.ts#L128

Is this something we need to extend it to support, or should I do this another way? (I'm not sure whether this
 pattern syntax is well-specified or we can just make the rules up, though I presume there's benefit in being aligned with eslint or VS Code's implementation where possible).

dbaeumer:
I try to stay away from Regexp in LSP (we have them for snippets but because they are in Textmate). If we start to use them we need to spec which features we support in RegExp in LSP. Glob patterns are well specified.

Actually the eslint matcher only needs to match directories no files. So it might not be a good candidate to use. We have one in VSCode here: https://github.com/microsoft/vscode/blob/master/src/vs/base/common/glob.ts#L446

May be we should just use https://www.npmjs.com/package/glob.

DanTup:
> Actually the eslint matcher only needs to match directories no files

Ah, that makes sense. I had started to modify it to support what I needed (including full nesting for the braces), though I'll take a look through the VS Code code to see if that's more easily reusable and does everything required (and if not, I'll try glob). Thanks!

DanTup:
@dbaeumer 

I was unable to make the VS Code matching or minimatch (which glob uses) behave as I believe we wanted (trailing slashes deciding whether to include only files or only folders), however I (think I) was able to get your es-lint code working with some changes:

https://github.com/DanTup/vscode-languageserver-node/commit/d494cd079db1cb3ff497301a7accdc8ee0be1b5e

I'll proceed using that for now, but if you think we should do it another way let me know.

> The middle ware uses VS Code types for requests going from the client to the server and LSP types for things flowing from the server to the client. This allows incepting the data and modifying it before it is processed.

Let's assume a server says it wants `{**/,}*.dart` (all folders and all dart files at any level) and a user drags `foo.dart` and `foo.txt` in the same action. Should we send the whole event to the server (since one of the files matched) or filter it only to those that matched. If the latter (as I expect), should the middleware get the full list or the filtered list?

Looking at the WillSaveWaitUntil, the middleware is called only if the original event matches the selectors, and any changes it made go straight to the server (so if the middleware changed the document details, it would still go to the server). To be consistent with this, I think we would filter the event before sending it to the middleware, then always send it on to the server (though I don't know if it's strange to filter the files in the event for the middleware?).

DanTup:
@dbaeumer I've opened a PR at https://github.com/microsoft/vscode-languageserver-node/pull/687 as a better place to discus the code. It's very incomplete, but I have questions about how best to test so looking for some pointers. Thanks!

dbaeumer (approved):

PR #1082, opened 2020-08-31
Remote Inspector page until compliance errors fixed

gregvanl:
Added a redirect back to overview page for any hard-coded URLs


dbaeumer:
Thanks!

PR #1091, opened 2020-09-08
Fix XML (LemMinX) implementation link

rzgry:
Noticed the link was broken for lsp4xml. lsp4xml was moved to eclipse and renamed to LemMinX.

FYI @angelozerr


angelozerr:
+1 Thanks @rzgry !

angelozerr:
Thanks @dbaeumer !

PR #1095, opened 2020-09-16
Update 3.16 spec to include moniker request method

BertanAygun:


dbaeumer:
Thanks again!

kjeremy:
Can this be backed out and reapplied without the line ending changes?

dbaeumer:
@kjeremy is this causing problems. I would actually prefer LF over CRLF since we use LF in the implementation project as well.

kjeremy:
No it's all good.

dbaeumer:
Thanks!

PR #1098, opened 2020-09-26
Add LspTypes for C# and clean up table.

kaby76:
I am adding here my LspTypes C# library, which implements every message parameter type in the 3.16 spec and works with VS2019, VSCode, Emacs lsp-emacs with my server [Antlrvsix](https://github.com/kaby76/AntlrVSIX). Other versions will be scraped directly from the online spec with a tool I will be writing because it took me a week to methodically hand-translate the spec into C#.

I cleaned up the table slightly by sorting it by the first column, which was done in the [servers table](https://microsoft.github.io/language-server-protocol/implementors/servers/). However, the link for "C# | MS | work in progress by David Wilson" is completely useless for people looking for resources to *actually implement code*. I suggest this row be deleted, or MS actually open-source whatever library that is referred to here.

### DIFF
--- a/_implementors/sdks.md
+++ b/_implementors/sdks.md
@@ -10,22 +10,23 @@ index: 3
 
 | Language | Maintainer | Repository |
 |------|--------|----------|
-| node.js | MS | [vscode-languageserver-node](https://github.com/Microsoft/vscode-languageserver-node)  |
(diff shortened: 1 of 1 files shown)


kaby76:
Thanks! It was a lot of work making the library. Hopefully, others can find it useful. --Ken

PR #1103, opened 2020-10-11
Update to new Microsoft cookie banner

gregvanl:
The only cookie currently is the GA essential cookie so we don't need to do any conditional creation or disabling.
This PR just updates the cookie banner and adds the "Manage cookies" link to the footer.

### DIFF
--- a/js/page.js
+++ b/js/page.js
@@ -4,6 +4,12 @@ $('#small-nav-dropdown').change(function() {
     .val()
 })
 
+function onConsentChanged() {
+  console.log(WcpConsent.siteConsent.getConsent());
+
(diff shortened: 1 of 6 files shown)


js/page.js
```diff
@@ -4,6 +4,12 @@ $('#small-nav-dropdown').change(function() {
...
 })
 
+function onConsentChanged() {
+  console.log(WcpConsent.siteConsent.getConsent());
+
+  // Where we would handle non-essential cookies in the future
+}
+
 $(function() {
   // Load GA upfront because we classify it as essential cookie
```
sbatten, line 14: what makes this an essential cookie?

PR #1116, opened 2020-10-27
[DISCUSSION]: Add dumpHash to LSIF Meta Data vertex

jepetty:
My proposal details why we'd like this, at least in a 1-off contract with LSIF providers for some languages that are common in extremely large repositories (i.e. C#). Is there interest/need in including this in the broader LSIF spec for utilization by more tools in a more standard way?

I appreciate thoughts and suggestions!


projectHashProposal.md
```diff
@@ -0,0 +1,18 @@
+# LSIF: ProjectWithHash
+
+Formulating and processing LSIF for large repositories can be very time consuming. We've discovered that many commits touch only ~20% of projects in a given repository, so there can be lots of unnecessary processing done on unchanged projects that are then ignored at ingestion time. To decrease processing times, I propose adding a new "projectHash" property to the existing [project vertex](https://github.com/Microsoft/language-server-protocol/blob/master/indexFormat/specification.md#the-project-vertex). The new definition for the `project` vertex would be:
+
+```ts
+export interface ProjectWithHash extends Project {
```
BertanAygun, line 6: Why not just make it part of the Project interface since it is optional already.

projectHashProposal.md
```diff
@@ -0,0 +1,18 @@
...
+    /**
+     * An optional hash of the files included in the project, `base64` encoded.
```
BertanAygun, line 8: I wouldn't mention files here since in most cases hash has to be calculated from input source files and references and may also depend on build configuration for example for C++ projects.

We should clarify that projectHash should only be same for 2 projects if their LSIF output is equivalent. I think the explanation below makes that appearent but it should be part of the property comment as well.

BertanAygun (changes requested):

projectHashProposal.md
```diff
@@ -0,0 +1,18 @@
...
+     */
+    projectHash?: string;
+}
+```
+
+Individual languages can decide what is the most reasonable selection for this project hash. For C#, this could be the MVID of the project assembly. For typescript, this could be a hash of the contents of all files included in the tsconfig.json. This `projectHash` property can then be compared against an existing cache of project hashes for the given repository to determine if the content has changed and the project needs to be re-ingested.
+
+# Rich Code Navigation: Application
```
BertanAygun, line 16: We can generalize this statement for indexers in general and not make it specific to Rich Navigation.

BertanAygun (changes requested):

projectHashProposal.md
jepetty: Great suggestion! Done

projectHashProposal.md
jepetty: Sounds good! I added the optional projectHash to the existing LSIF spec, but kept this doc for now to facilitate discussion - I don't know if we want this much detail to enter the LSIF spec for this single property.

indexFormat/specification.md
```diff
@@ -807,7 +807,13 @@ export interface Project extends V {
 	/**
 	 * Optional the content of the project file, `base64` encoded.
 	 */
-	contents?: string;
+  contents?: string;
+
+  /**
+   * An optional hash of the files included in the project, `base64` encoded. projectHash should be the same
```
jepetty, line 813: No need to define the encoding or any specifications for this string - opaque string that language determines. Specify size - 512 chars maybe?

dbaeumer:
@jepetty any objection to close. I added the field as we have discussed offline.

jepetty:
@dbaeumer sorry, I should have closed it earlier! Yes, I'll close now :) Thank you!

Issue #1117, opened 2020-10-28
Provide a client-side protocol message that allows server to request execution of a command


kdvolder:
There is existing protocol for a client to request a server to execute a command. We have several places where we want to have the reverse. The server asks the client to execute command instead. 

This would work in the same way as the client->server message: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_executeCommand 

We could use the same name `workspace/executeCommand` or use a different name to emphasize that the message flows in the opposite direction e.g. `workspace/clientExecuteCommand`.

This request may seem a bit strange if we assume that all commands are defined by the server itself. However that is not typically the case in practice. In practice the server contributes commands and these commands get entered into some type of client-side command registry. There are actually more commands that exist client-side that were not defined by the server itself. (e.g. some commands may be defined by the client itself, or some commands may defined by other servers with which the current server interoperates somehow).

For this proposal to make sens, all we have to assume is that commands can 'exist' on the client that are not defined by the server itself. We do not need to assume a specific mechanism by which these commands are defined. It should be possible to agree on a 'standard' mechanism to request the execution of a client-side command.

For some context...

We have run into several case (3) where we needed this and in each case it was handled by some custom protocol extension.

The latest of these is here: https://github.com/eclipse/wildwebdeveloper/pull/543#issuecomment-714611151

There has also been some discussion in other related LSP tickets, which I can sumarize here:

- #430
- #432
- #642

The first one is probably the most closely related, one part of what it needs/asks for is a mechanism for server to request client to execute a command.

The second one has a discussion around whether commands live in a 'shared/global' name space or whether the commands belong only to a specific client<->server session.

With a shared namespace one server could define a command and another one could use it (e.g. in code action, or also in a potential executeClientCommand request. With 'per session namespace' this wouldn't be possible.

These tickets digressed into long discussion which do not seem immediately actionable.

So I propose in this ticket to narrow the scope a bit, to make it more actionable.

Let's avoid getting into discussions about:

- how exactly commands may be defined on the client side (we just need to accept that they can be defined/exist *somehow*)
- whether client-side commands are in a global name space
- whether/how a server can discover which commands are supported on the client

Instead I intend this ticket to be **only about adding a single protocol message that allows the server to request the client to execute a command**.

dbaeumer:
@kdvolder would you be willing to work on a PR for this. Both for the spec and the vscode-languageserver-node implementation?

mickaelistria:
[...]

PR #1119, opened 2020-10-29
client/executeCommand

kdvolder:
See: https://github.com/microsoft/language-server-protocol/issues/1117

### DIFF
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -2804,6 +2804,11 @@ _Response_:
 * result: `any` \| `null`
 * error: code and message set in case an exception happens during the request.
 
+The client shall return `ErrorCode.InvalidParams` when the server requests
+
+- a command that doesn't exist
+- invalid params for the command signature
+
 #### <a href="#workspace_applyEdit" name="workspace_applyEdit" class="anchor">Applies a WorkspaceEdit (:arrow_right_hook:)</a>
 
 The `workspace/applyEdit` request is sent from the server to the client to modify resource on the client side.
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -2715,7 +2715,7 @@ _Response_:
 * partial result: `SymbolInformation[]` as defined above.
 * error: code and message set in case an exception happens during the workspace symbol request.
 
-#### <a href="#workspace_executeCommand" name="workspace_executeCommand" class="anchor">Execute a command (:leftwards_arrow_with_hook:)</a>
+#### <a href="#workspace_executeCommand" name="workspace_executeCommand" class="anchor">Execute a server command (:leftwards_arrow_with_hook:)</a>
 
 The `workspace/executeCommand` request is sent from the client to the server to trigger command execution on the server. In most cases the server creates a `WorkspaceEdit` structure and applies the changes to the workspace using the request `workspace/applyEdit` which is sent from the server to the client.
 
@@ -2774,6 +2774,31 @@ export interface ExecuteCommandParams extends WorkDoneProgressParams {
 ```
 
 The arguments are typically specified when a command is returned from the server to the client. Example requests that return a command are `textDocument/codeAction` or `textDocument/codeLens`.
+
+_Response_:
+* result: `any` \| `null`
+* error: code and message set in case an exception happens during the request.
+
+#### <a href="#client_executeCommand" name="client_executeCommand" class="anchor">Execute a client command(:leftwards_arrow_with_hook:)</a>
+
+The `client/executeCommand` request is similar to `workspace/executeCommand` but is sent from the server to the client, to trigger command execution on the client. 
+
+_Client Capability_:
+* property path (optional): `client.executeCommand`
+* property type: `ClientExecuteCommandClientCapabilities` defined as follows:
+
+```typescript
+export interface ClientExecuteCommandClientCapabilities {
+	/**
+	 * The client supports message 'client/executeCommand'.
+	 */
+	supported: boolean;
+}
+```
+
+_Request:_
+* method: 'client/executeCommand'
+* params: `ExecuteCommandParams` defined exactly the same way as for 'workspace/executeCommand'. 
 
 _Response_:
 * result: `any` \| `null`


_specifications/specification-3-16.md
```diff
@@ -2779,6 +2779,31 @@ _Response_:
...
+_Client Capability_:
+* property path (optional): `client.executeCommand`
+* property type: `ClientExecuteCommandClientCapabilities` defined as follows:
+
+```typescript
+export interface ClientExecuteCommandClientCapabilities {
+	/**
+	 * The client supports message 'client/executeCommand'.
+	 */
+	supported: boolean;
```
rwols, line 2795: I get that you're trying get something out the door but this needs many more options for clients. Given just this state of the PR a server doesn't know whether a command can succeed or not.

Just to give you some ideas:

- the client should be able to communicate dynamically to the server about possible new commands that it knows about, similar to workspace/didChangeConfiguration.
- It should be possible for the server to get a list of known command names from the client, similar to workspace/configuration

At this point the server can only do client/executeCommand and hope for the best.

rwols:
If this ultimately also leads to standardizing commands like "run auto-complete" and "commit completion" in a client-neutral way then I'm okay with this.

_specifications/specification-3-16.md
```diff
@@ -2779,6 +2779,31 @@ _Response_:
...
+}
+```
+
+_Request:_
+* method: 'client/executeCommand'
+* params: `ExecuteCommandParams` defined exactly the same way as for 'workspace/executeCommand'. 
+
+_Response_:
+* result: `any` \| `null`
+* error: code and message set in case an exception happens during the request.
```
rwols, line 2805: I would suggest extra wording here:

The client shall return ErrorCode.InvalidParams when the server requests

- a command that doesn't exist
- invalid params for the command signature

_specifications/specification-3-16.md
mickaelistria: I don't think we should require more options at this point. The current proposal is good and there is room for evolution (such as listing commands) in the future. However, we don't really have a strong need to pass a static list of commands at that point, just like symmetrically, there is no strong need for clients to know in advance the list of server commands it can invoke.
The command pattern is often use for client or LS specific integrations, that we cannot really standardize in LSP because they're not generic enough.
I think it's better to mimic the current `workspace/executeCommand` action (and related capabilities), it seems more consistent.

> At this point the server can only do client/executeCommand and hope for the best.

It's the same for `workspace/executeCommand`, and yet this is working well ;)

mickaelistria:
> If this ultimately also leads to standardizing commands like "run auto-complete" and "commit completion" in a client-neutral way then I'm okay with this.

I don't get the use-case here, the client is supposed to orchestrate that directly with LSP operations, why would command be useful for that?
If we can identify some generally useful server side commands that we're ready to spec, then it's better to just specify them as proper LSP operation, with proper parameters definition and so on than leaving it in a pot-pourri of command ids without finer specification.

_specifications/specification-3-16.md
rwols: > It's the same for workspace/executeCommand, and yet this is working well ;)

Not quite. If you're lucky, you can read a readme file explaining what the arguments of a command can be. If you're unlucky, you have to dig through the server source code to figure out what a command can do.

rwols: The configure-less way of invoking commands is via code actions, which works well, I agree.

rwols:
> I don't get the use-case here, the client is supposed to orchestrate that directly with LSP operations, why would command be useful for that?

There are servers that put `"command": "editor.action.triggerSuggest"` in some of their completion items. This means "show the auto-complete widget". But it's not standardized.

> If we can identify some generally useful server side commands that we're ready to spec, then it's better to just specify them as proper LSP operation, with proper parameters definition and so on than leaving it in a pot-pourri of command ids without finer specification.

For sure :)

_specifications/specification-3-16.md
kdvolder: I'm not against adding more options for things such as 'command discovery' or making the spec more precise about mechanism by which commands are made to 'exist' on the client side.

However, I wanted to explicitly avoid getting this become yet another ticket dragged out into a lengthy discussion with no actionable outcome. 

To me the questions:

1. how does a server request a client to execute a command
2. how does a server know that a client can execute a specific (set of) commands
3. how are client-side commands defined

So I explicitly wanted limit the scope of this ticket to be just about 1. The other questions can be answered independently and in fact there are already some tickets open with long discussions.

> Given just this state of the PR a server doesn't know whether a command can succeed or not.

That is true, but at least we know how to request execution in a standardised way. To me that is progress. 

And we can assume an error comes back if the execution failed (I agree with your other comments that we can perhaps make some of the error cases more explicit in the spec).

_specifications/specification-3-16.md
kdvolder: Agreed. Done!

_specifications/specification-3-16.md
rwols: No worries. I am just some guy maintaining a client and weary of opening pandora's box with client/executeCommand. We must have a follow-up PR that defines how command discovery works, and we must standardize some "common" commands.

_specifications/specification-3-16.md
```diff
@@ -2779,6 +2779,36 @@ _Response_:
 * result: `any` \| `null`
 * error: code and message set in case an exception happens during the request.
 
+#### <a href="#client_executeCommand" name="client_executeCommand" class="anchor">Execute a client command(:leftwards_arrow_with_hook:)</a>
+
+The `client/executeCommand` request is similar to `workspace/executeCommand` but is sent from the server to the client, to trigger command execution on the client. 
+
+_Client Capability_:
+* property path (optional): `client.executeCommand`
```
rwols, line 2787: This is the first client property path that starts with `client.`. Why not `workspace.`?

Should we be able to distinguish between workspace and textDocument commands?

kdvolder: I would just make the property mirror the name of the message. That seems to be the pattern so far. So, if we call the message 'client/executeCommand' than I think we should call the property 'client.executeCommand'. 

But I personally don't care all that much about what exactly we call either the message or the property. 

So I'm okay if you suggest some other names for both and I'll just go along with whatever other people seem to like best.

If you want to suggest different name (for message or property), it would be best to suggest the full names you'd prefer (`workspace.` is only the prefix and I think we can't use `workspace.executeCommand` because that's already used for the client->server message).

rwols: The only other client/ requests seem to be client/registerCapability and client/unregisterCapability. But those don't have a client capability. I'm fine with `client.executeCommand` :)

kdvolder: So does that mean we can resolve the conversation? It sounds like you are fine leaving it as is? But I'm not sure that's what you are really saying :-).

rwols: Yes, I meant that you can resolve the convo. Sorry for the confusion!

rwols (approved):

_specifications/specification-3-16.md
kdvolder: I don't really disagree with you. I just don't want to go there yet.  Yes, I could imagine some options added to capability object in the future such as 'commands: [...]' for a list of 'statically' known commands. Or a property 'commandDisovery: boolean' that indicates there is some extra protocol for command discovery. etc. But this sort of thing could be arbitrarily complex and with lots of people wanting something a little different from it. So I just don't think we should/can tackle all of these at once right now. The more we can separate this into smaller questions/issues, the more chance we have of actually making progress in small increments.

mickaelistria (approved):

dbaeumer:
@kdvolder thanks for the PR. Besides updating the specification we need a reference implementation somewhere as well. This can be VS Code, Eclipse or any other client. Is someone willing to do that as well?

kdvolder:
> This can be VS Code, Eclipse or any other client. Is someone willing to do that as well?

I was planning on doing it yes, just need to find the time :-)

dbaeumer:
@kdvolder great.

radeksimko:
I appreciate the work being done here and aim to ship something.

Do you mind describing some concrete use cases this is aiming to solve?

I ran into at least two use cases myself where I had the need to execute client-side commands:
 - open completion prompt
 - show references

In both cases these are commands that would benefit from being linked directly from their relevant methods - i.e. [`CompletionItem.command`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionItem) or [`CodeLens.command`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#codeLens).

Executing a separate standalone method would require either extra roundtrip where the server implements some kind of a proxy server-side command for any client-side command, or somehow entirely bypass the existing workflow.

Are there some other use cases this is meant to address?

--- 

I agree with @rwols here that for server to not know what commands are available on the client it makes it practically impossible to provide a decent UX in cases I outlined above. Imagine that server implements code lens to show reference counts, user clicks to "5 references", linking to `editor.action.showReferences` and the client passes it back to the server because it doesn't implement it. The server can't do anything about it so the click ends up as no-op in the UI. That doesn't seem ideal? It would seem better if the server could just suppress the code lens entirely and avoid showing the link in the first place.

radeksimko:
I just found a use case for this feature. I assume it would be useful for updating some UI elements when the server deems it necessary.

The client may pull certain details (such as language/dependency versions or constraints, or some sort of hierarchy) about the codebase via custom command, but these details may get outdated as the user works on the codebase. It would therefore be useful for the server to notify the client when such details change.

I suppose that is the reason why these `refresh` commands exist anyway

 - [`workspace/codeLens/refresh`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#codeLens_refresh)
 - [`workspace/semanticTokens/refresh`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#semanticTokens_refreshRequest)

shreyashHake (approved):

dbaeumer:
Any progress on this PR?

dbaeumer:
This issue has been closed because it needs more information and has not had recent activity. See also our [issue reporting](https://aka.ms/vscodeissuereporting) guidelines.

Happy Coding!

PR #1123, opened 2020-11-03
Escaping dollar signs

AndreasMatthias:
Pairwise dollar signs on the same line are misinterpreted as math delimiters if they are not escaped.


AndreasMatthias:
This PR was a mistake. I'm sorry.

PR #1124, opened 2020-11-03
Escape dollar signs.

AndreasMatthias:
Pairwise dollar signs on the same line are misinterpreted as math delimiters if they are not escaped.


dbaeumer:
Thanks!

PR #1134, opened 2020-11-07
`DocumentSymbol.detail?`: support `MarkupContent`

### LINKED ISSUE #1130: `DocumentSymbol.detail?`: syntax highlighting
The `DocumentSymbol` interface exposes a `detail?` field often contains signatures of functions. This is very useful when implementing outline views, as additional information can be provided to the user (the symbol's signature instead of only its name).

However, because `DocumentSymbol.detail?` is just a string, the client cannot properly colorize the output the same way as is possible for `textDocument/signatureHelp`. I propose that the `textDocument/documentSymbol` request be extended with a paramter that makes the `DocumentSymbol.detail?` of its results `MarkupContent`, and perhaps even allows to control whether that field should be present at all. A client capability for this is probably the wrong approach, as there are many uses for `textDocument/documentSymbol`, and not all of them require the more heavy markdown.

Also, perhaps `textDocument/documentSymbol` should allow lazily requesting the children, similar to the "variables" request of the Debugger Adapter Protocol?

Thoughts and opinions welcome.

nbfalcon:
Fixes #1130.

### DIFF
--- a/_specifications/specification-3-15.md
+++ b/_specifications/specification-3-15.md
@@ -4441,11 +4441,15 @@ export interface DocumentSymbolClientCapabilities {
 		 */
 		valueSet?: SymbolKind[];
 	}
-
 	/**
 	 * The client supports hierarchical document symbols.
 	 */
 	hierarchicalDocumentSymbolSupport?: boolean;
+	
+	/**
+	 * The client supports `MarkupContent` in `DocumentSymbol.detail?`.
+	 */
+	detailMarkupSupport?: boolean;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


KamasamaK:
LSP v3.15 was already finalized. Any new changes would need to be done to `specification-3-16.md`.

nbfalcon:
Fixed.

rwols:
It feels a bit odd because for hover and completion docs the client can advertise a list of supported formats (currently: `plaintext` and/or `markdown`). Should this diff follow that principle?

nbfalcon:
Actually, that is a good idea :); I'll fix this soon.

nbfalcon:
On second thought, shouldn't we perhaps have a shared list for the various supported formats? Why would a client be able to render Markdown for auto-completion, but not for signatureHelp or this field?

rwols:
A client may decide to only allow plaintext for the detail field for performance reasons :)

dbaeumer:
@nbfalcon thanks for the PR. Just want to point out that besides a spec update we need a proof of concept implementation as well. This doesn't need to be done in VS Code. It can be any editor that supports LSP. See https://github.com/microsoft/language-server-protocol/blob/master/contributing.md

Are you willing to work on this as well?

nbfalcon:
I wanted to do that for Emacs `lsp-mode` anyway. The necessary refactoring is already done.

dbaeumer:
Great. Can you post a screen shot here on how this will look?

dbaeumer:
Ant update on an implementation?

dbaeumer:
Closing due to inactivity.

Issue #1087, opened 2020-09-05
Rename 'member' semantic token type


0dinD:
I implemented redhat-developer/vscode-java#1488 a while ago, and was very confused as to why aeschli wanted to use the `member` semantic token type instead of `function` for Java methods. As it turns out (took me a while to figure out), the name `member` is misleading since the token type actually refers to ***member functions***.

This is also misleading for theme authors, as you would expect to style ***all members*** (including member variables, for example) using the `member` token type. But in reality, `member` only applies to member functions, not member variables. 

Is it too much trouble to rename the `member` token type to `memberFunction`?  
If it is, I think a note in the documentation is a good idea in order to prevent future misunderstandings.

rcjsuen:
@0dinD If `member` is for member functions then how do I style member variables? 🤔

0dinD:
@rcjsuen that's done using the semantic token type `property` (as far as I can tell), which now that I think about it is also a bit confusing. I still think `member` is the most confusing and misleading token type out of the two, but I wouldn't complain if `property` was renamed to `memberVariable` while we're at it.

The meaning of the semantic token types are not very well documented in VS Code or LSP, but from [this line](https://github.com/microsoft/vscode/blob/de7bb5b5fe55fe562361df6eeed61d970929c936/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L525) and [this line](https://github.com/microsoft/vscode/blob/de7bb5b5fe55fe562361df6eeed61d970929c936/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L530) you can tell what the token types are supposed to represent based on the assigned TextMate scopes.

dbaeumer:
@aeschli can you please comment / help me.

dbaeumer:
@aeschli ping!

sam-mccall:
I got bitten by this too (https://github.com/clangd/vscode-clangd/issues/105)

#1146 would make the status quo clearer in the spec at least.
Using clearer names would be nice though!

dbaeumer:
@aeschli thanks. I will adopt in LSP.

PR #1146, opened 2020-11-20
Clarify SemanticTokenTypes.member is for functions

sam-mccall:
Many languages including C and C++ define "member" to include consider instance variables/fields/data members.
However VSCode maps `member` to the textmate scope `entity.name.function.member`, i.e. method/member function.
The intent seems to be that methods are `member` and fields are `property`.

It may be too late to change the names to be less ambiguous, but at least we can clarify in the spec.

### DIFF
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -6980,6 +6980,7 @@ export enum SemanticTokenTypes {
 	enumMember = 'enumMember',
 	event = 'event',
 	function = 'function',
+	/** A member function, or method. */
 	member = 'member',
 	macro = 'macro',
 	keyword = 'keyword',


rwols:
duplicate of #1087 ?

sam-mccall:
Indeed, thanks!
(Might still be worth adding the comment while waiting for a better solution)

dbaeumer:
@aeschli can you have a look and let me know if this assumption is correct.

dbaeumer:
I will close the PR. We discussed this internally and since the current state is still not final we decided to rename member -> method.

sam-mccall:
Thanks Dirk, the rename is even better.

PR #1152, opened 2020-11-25
Support readonly Location/LocationLink and showDocument

### LINKED ISSUE #1150: Specify read-only locations
It would be nice to provide a hint that locations (`Location`, `LocationLink`) could be read-only. For instance: if performing a find usages or a go to definition on a symbol that ends up in a third party library (crates, npm packages, disassembled .NET DLLs) you may not want that source to be editable at that location and could open up a document in read-only mode.

https://github.com/rust-analyzer/rust-analyzer/issues/6623

kjeremy:
Allows the server to specify that the client should display a location in a read only mode. The client should advertise this support using `general.readonlySupport`.

Fixes #1150

### DIFF
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -433,10 +433,16 @@ interface Range {
 #### <a href="#location" name="location" class="anchor"> Location </a>
 
 Represents a location inside a resource, such as a line inside a text file.
+The server may optionally hint to the client that a location should be considered readonly.
 ```typescript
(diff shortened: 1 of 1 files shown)


dbaeumer:
@kjeremy thanks. But please keep in mind that we need a client implementation as well.

dbaeumer:
Any progress on a client implementation?

dbaeumer:
Closing due to inactivity.

PR #1156, opened 2020-12-01
Tweaks to simplify parsing of spec

DanTup:
@dbaeumer would you accept these tweaks to improve parsing the spec? One of the definitions uses `URI` as a type, but it's not defined in code anywhere. My understanding is that it's a string (same as `DocumentUri`).

The other change makes the type of a *response* a single block of code, so it can be more easily identified. Multiple code blocks is a little more tricky to parse and I don't think anything is lost doing it this way.


DanTup:
Thanks!

PR #1161, opened 2020-12-03
Clarify window/workDoneProgress/cancel usage

### LINKED ISSUE #1159: Can the client issue window/workDoneProgress/cancel for any progress token?
Can we expect the client to be able to cancel any tokens (for instance if there's an error) created with `window/workDoneProgress/create` or only those marked as `cancellable` in `WorkDoneProgressBegin` or `WorkDoneProgressReport`?

kjeremy:
Fixes #1159


dbaeumer:
Appreciated!

PR #1167, opened 2020-12-14
Remove duplicate URI definition

DanTup:
@dbaeumer spotted this when re-generating my classes. This was added in both https://github.com/microsoft/language-server-protocol/pull/1156 and bae2aaba5946c84cb28c28f4437711d744576ed7 so there were two. This removes the one added in my PR and leaves just yours.


dbaeumer:
Thanks

PR #1168, opened 2020-12-14
Remove Go SDK

disposedtrolley:
I added an entry to the `SDKs for the LSP` page a long time ago with my attempt at an LSP SDK written in Go. Unfortunately I haven't had much time to implement it, so I thought it'd be a good idea to remove it from the list as to not mislead anybody.

Cheers!


dbaeumer:
Thanks for getting back.

PR #1169, opened 2020-12-16
Fix case of textEdit link

rchl:
Incorrect:
https://microsoft.github.io/language-server-protocol/specification#textedit

Correct:
https://microsoft.github.io/language-server-protocol/specification#textEdit

Also added missing parenthesis in some old version of the spec.


dbaeumer:
Thanks!

PR #1231, opened 2021-03-29
Support Type Hierarchy

CsCherrYY:
There are three requests to support type hierarchy:
- `textDocument/prepareTypeHierarchy`: Sent from the client to the server to get the type hierarchy item from the given text document position
- `typeHierarchy/supertypes` and `typeHierarchy/subtypes`: Sent from the client to the server to resolve the given item's supertypes or subtypes

### DIFF
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -8302,7 +8302,7 @@ export interface TypeHierarchyItem {
 	 * Indicates if this item has subtypes. When `undefined` the subtypes
 	 * of this item have not been resolved yet.
 	 */
-	hasSubtypes?: boolean
+	hasSubtypes?: boolean;
 
 	/**
 	 * A data entry field that is preserved between a call hierarchy prepare and
@@ -8368,11 +8368,11 @@ The request is sent from the client to the server to resolve the inheritance tre
 * `TreeItem<T>` defined as follows:
 ```typescript
 export interface TreeItem<T> {
-	data: T,
+	data: T;
 	/**
 	 * The children of this TreeItem. When `undefined` the children have not been resolved yet.
 	 */
-	children?: TreeItem<T>[],
+	children?: TreeItem<T>[];
 }
 ```
 The result represents the inheritance tree related to the specified type hierarchy item. Only single inheritance can be represented this way. The root type is expected as the return value. The request doesn't define its own client and server capabilities. It is only issued if a server has the capability for `TypeHierarchyOptions/inheritanceTreeSuppport`.
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -8197,8 +8197,6 @@ The type hierarchy request is sent from the client to the server to return a typ
   1. first a type hierarchy item is prepared for the given text document position.
   1. for a type hierarchy item the supertype or subtype type hierarchy items are resolved.
 
-In the first step, the `textDocument/prepareTypeHierarchy` request could have a unique, constant and optional `transactionId`. The following `typeHierarchy/supertypes` and `typeHierarchy/subtypes` requests in the second step could have the same `transactionId` in their params, which could be used to help indicate some cached data in the server.
-
 _Client Capability_:
 
 * property name (optional): `textDocument.typeHierarchy`
@@ -8223,6 +8221,10 @@ _Server Capability_:
 
 ```typescript
 export interface TypeHierarchyOptions extends WorkDoneProgressOptions {
+	/** 
+	 * The server supports for providing an inheritance tree.
+	 */
+	inheritanceTreeSuppport?: boolean;
 }
 ```
 
@@ -8243,7 +8245,6 @@ _Request_:
 ```typescript
 export interface TypeHierarchyPrepareParams extends TextDocumentPositionParams,
 	WorkDoneProgressParams {
-	transactionId?: integer | string;
 }
 ```
 
@@ -8293,7 +8294,9 @@ export interface TypeHierarchyItem {
 
 	/**
 	 * A data entry field that is preserved between a type hierarchy prepare and
-	 * supertypes or subtypes requests.
+	 * supertypes or subtypes requests. It could also be used to identify the 
+	 * type hierarchy in the server, helping improve the performance on 
+	 * resolving supertypes and subtypes.
 	 */
 	data?: unknown;
 }
@@ -8315,8 +8318,11 @@ _Request_:
 ```typescript
 export interface TypeHierarchySupertypesParams extends
 	WorkDoneProgressParams, PartialResultParams {
+	/**
+	 * If this is set to `true`, The request only asks for super class.
+	 */
+	classOnly?: boolean;
 	item: TypeHierarchyItem;
-	transactionId?: integer | string;
 }
 ```
 _Response_:
@@ -8340,7 +8346,6 @@ _Request_:
 export interface TypeHierarchySubtypesParams extends
 	WorkDoneProgressParams, PartialResultParams {
 	item: TypeHierarchyItem;
-	transactionId?: integer | string;
 }
 ```
 _Response_:
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -8194,8 +8194,8 @@ export interface Moniker {
 
 The type hierarchy request is sent from the client to the server to return a type hierarchy for the language element of given text document positions. Will return `null` if the server couldn't infer a valid type from the position. The type hierarchy requests are executed in two steps:
 
-  1. first a type hierarchy item is prepared for the given text document position
-  1. the client request the server to resolve the given item with its supertypes or subtypes
+  1. first a type hierarchy item is prepared for the given text document position.
+  1. for a type hierarchy item the supertype or subtype type hierarchy items are resolved.
 
 In the first step, the `textDocument/prepareTypeHierarchy` request could have a unique, constant and optional `transactionId`. The following `typeHierarchy/supertypes` and `typeHierarchy/subtypes` requests in the second step could have the same `transactionId` in their params, which could be used to help indicate some cached data in the server.
 
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -1927,6 +1927,13 @@ export interface TextDocumentClientCapabilities {
 	 * @since 3.16.0
 	 */
 	moniker?: MonikerClientCapabilities;
+
+	/**
+	 * Capabilities specific to the various type hierarchy requests.
+	 *
+	 * @since 3.17.0
+	 */
+	typeHierarchy?: TypeHierarchyClientCapabilities;
 }
 ```
 
@@ -2418,6 +2425,14 @@ interface ServerCapabilities {
 			willDelete?: FileOperationRegistrationOptions;
 		}
 	}
+
+	/**
+	 * The server provides type hierarchy support.
+	 *
+	 * @since 3.17.0
+	 */
+	typeHierarchyProvider?: boolean | TypeHierarchyOptions
+		| TypeHierarchyRegistrationOptions;
 
 	/**
 	 * Experimental server capabilities.
@@ -8172,6 +8187,211 @@ export interface Moniker {
 	kind?: MonikerKind;
 }
 ```
+
+#### <a href="#textDocument_prepareTypeHierarchy" name="textDocument_prepareTypeHierarchy" class="anchor">Prepare Type Hierarchy Request (:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The type hierarchy request is sent from the client to the server to return a type hierarchy for the language element of given text document positions. The type hierarchy requests are executed in two steps:
+
+  1. first a type hierarchy item is prepared for the given text document position
+  1. the client request the server to resolve the given item with its supertypes or/and subtypes
+
+_Client Capability_:
+
+* property name (optional): `textDocument.typeHierarchy`
+* property type: `TypeHierarchyClientCapabilities` defined as follows:
+
+```typescript
+interface TypeHierarchyClientCapabilities {
+	/**
+	 * Whether implementation supports dynamic registration. If this is set to
+	 * `true` the client supports the new `(TextDocumentRegistrationOptions &
+	 * StaticRegistrationOptions)` return value for the corresponding server
+	 * capability as well.
+	 */
+	dynamicRegistration?: boolean;
+}
+```
+
+_Server Capability_:
+
+* property name (optional): `typeHierarchyProvider`
+* property type: `boolean | TypeHierarchyOptions | TypeHierarchyRegistrationOptions` where `TypeHierarchyOptions` is defined as follows:
+
+```typescript
+export interface TypeHierarchyOptions extends WorkDoneProgressOptions {
+	/**
+	 * The server has support for supporting inheritance tree.
+	 */
+	inheritanceTreeSuppport?: boolean;
+}
+```
+
+_Registration Options_: `TypeHierarchyRegistrationOptions` defined as follows:
+
+```typescript
+export interface TypeHierarchyRegistrationOptions extends
+	TextDocumentRegistrationOptions, TypeHierarchyOptions,
+	StaticRegistrationOptions {
+}
+```
+
+_Request_:
+
+* method: 'textDocument/prepareTypeHierarchy'
+* params: `TypeHierarchyPrepareParams` defined as follows:
+
+```typescript
+export interface TypeHierarchyPrepareParams extends TextDocumentPositionParams,
+	WorkDoneProgressParams {
+}
+```
+
+_Response_:
+
+* result: `TypeHierarchyItem[] | null` defined as follows:
+
+```typescript
+export interface TypeHierarchyItem {
+	/**
+	 * The name of this item.
+	 */
+	name: string;
+
+	/**
+	 * The kind of this item.
+	 */
+	kind: SymbolKind;
+
+	/**
+	 * Tags for this item.
+	 */
+	tags?: SymbolTag[];
+
+	/**
+	 * More detail for this item, e.g. the signature of a function.
+	 */
+	detail?: string;
+
+	/**
+	 * The resource identifier of this item.
+	 */
+	uri: DocumentUri;
+
+	/**
+	 * The range enclosing this symbol not including leading/trailing whitespace
+	 * but everything else, e.g. comments and code.
+	 */
+	range?: Range;
+
+	/**
+	 * The range that should be selected and revealed when this symbol is being
+	 * picked, e.g. the name of a function. Must be contained by the
+	 * [`range`](#TypeHierarchyItem.range).
+	 */
+	selectionRange: Range;
+
+	/**
+	 * Indicates if this item has supertypes. When `undefined` the supertypes
+	 * of this item have not been resolved yet.
+	 */
+	hasSupertypes?: boolean;
+
+	/**
+	 * Indicates if this item has subtypes. When `undefined` the subtypes
+	 * of this item have not been resolved yet.
+	 */
+	hasSubtypes?: boolean
+
+	/**
+	 * A data entry field that is preserved between a call hierarchy prepare and
+	 * incoming calls or outgoing calls requests.
+	 */
+	data?: unknown;
+}
+```
+
+* error: code and message set in case an exception happens during the 'textDocument/prepareTypeHierarchy' request
+
+#### <a href="#typeHierarchy_supertypes" name="typeHierarchy_supertypes" class="anchor">Type Hierarchy Supertypes(:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The request is sent from the client to the server to resolve the supertypes for a given type hierarchy item. The request doesn't define its own client and server capabilities. It is only issued if a server registers for the [`textDocument/prepareTypeHierarchy` request](#textDocument_prepareTypeHierarchy).
+
+_Request_:
+
+* method: 'typeHierarchy/supertypes'
+* params: `TypeHierarchySupertypesParams` defined as follows:
+
+```typescript
+export interface TypeHierarchySupertypesParams extends
+	WorkDoneProgressParams, PartialResultParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TypeHierarchyItem[] | null`
+* partial result: `TypeHierarchyItem[]`
+* error: code and message set in case an exception happens during the 'typeHierarchy/supertypes' request
+
+#### <a href="#typeHierarchy_subtypes" name="typeHierarchy_subtypes" class="anchor">Type Hierarchy Subtypes(:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The request is sent from the client to the server to resolve the subtypes for a given type hierarchy item. The request doesn't define its own client and server capabilities. It is only issued if a server registers for the [`textDocument/prepareTypeHierarchy` request](#textDocument_prepareTypeHierarchy).
+
+_Request_:
+
+* method: 'typeHierarchy/subtypes'
+* params: `TypeHierarchySubtypesParams` defined as follows:
+
+```typescript
+export interface TypeHierarchySubtypesParams extends
+	WorkDoneProgressParams, PartialResultParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TypeHierarchyItem[] | null`
+* partial result: `TypeHierarchyItem[]`
+* error: code and message set in case an exception happens during the 'typeHierarchy/subtypes' request
+
+#### <a href="#typeHierarchy_inheritanceTree" name="typeHierarchy_inheritanceTree" class="anchor">Type Hierarchy Inheritance Tree(:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The request is sent from the client to the server to resolve the inheritance tree for a given type hierarchy item. The response result of this request is `TreeItem<T>`.
+* `TreeItem<T>` defined as follows:
+```typescript
+export interface TreeItem<T> {
+	data: T,
+	/**
+	 * The children of this TreeItem. When `undefined` the children have not been resolved yet.
+	 */
+	children?: TreeItem<T>[],
+}
+```
+The result represents the inheritance tree related to the specified type hierarchy item. Only single inheritance can be represented this way. The root type is expected as the return value. The request doesn't define its own client and server capabilities. It is only issued if a server has the capability for `TypeHierarchyOptions/inheritanceTreeSuppport`.
+
+_Request_:
+
+* method: 'typeHierarchy/inheritanceTree'
+* params: `TypeHierarchyInheritanceTreeParams` defined as follows:
+
+```typescript
+export interface TypeHierarchyInheritanceTreeParams extends
+	WorkDoneProgressParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TreeItem<TypeHierarchyItem> | null`
+* error: code and message set in case an exception happens during the 'typeHierarchy/resolve' request
 
 ##### Notes
 
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -8192,10 +8192,12 @@ export interface Moniker {
 
 > *Since version 3.17.0*
 
-The type hierarchy request is sent from the client to the server to return a type hierarchy for the language element of given text document positions. The type hierarchy requests are executed in two steps:
+The type hierarchy request is sent from the client to the server to return a type hierarchy for the language element of given text document positions. Will return `null` if the server couldn't infer a valid type from the position. The type hierarchy requests are executed in two steps:
 
   1. first a type hierarchy item is prepared for the given text document position
-  1. the client request the server to resolve the given item with its supertypes or/and subtypes
+  1. the client request the server to resolve the given item with its supertypes or subtypes
+
+In the first step, the `textDocument/prepareTypeHierarchy` request could have a unique, constant and optional `transactionId`. The following `typeHierarchy/supertypes` and `typeHierarchy/subtypes` requests in the second step could have the same `transactionId` in their params, which could be used to help indicate some cached data in the server.
 
 _Client Capability_:
 
@@ -8221,10 +8223,6 @@ _Server Capability_:
 
 ```typescript
 export interface TypeHierarchyOptions extends WorkDoneProgressOptions {
-	/**
-	 * The server has support for supporting inheritance tree.
-	 */
-	inheritanceTreeSuppport?: boolean;
 }
 ```
 
@@ -8245,6 +8243,7 @@ _Request_:
 ```typescript
 export interface TypeHierarchyPrepareParams extends TextDocumentPositionParams,
 	WorkDoneProgressParams {
+	transactionId?: integer | string;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


_specifications/specification-3-17.md
```diff
@@ -2419,6 +2426,14 @@ interface ServerCapabilities {
 		}
 	}
 
+	/**
+	 * The server provides type hierarchy support.
+	 *
+	 * @since 3.17.0
+	 */
+	typeHierarchyProvider?: boolean | TypeHierarchyOptions
```
KamasamaK, line 2434: Is a simple `boolean` necessary/sufficient here since each server capability can be defined under `TypeHierarchyOptions`?

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+```typescript
+export interface TypeHierarchyInheritanceTreeParams extends
+	WorkDoneProgressParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TreeItem<TypeHierarchyItem> | null`
+* error: code and message set in case an exception happens during the 'typeHierarchy/resolve' request
```
KamasamaK, line 8394: Is this supposed to be 'typeHierarchy/inheritanceTree'?

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+_Server Capability_:
+
+* property name (optional): `typeHierarchyProvider`
+* property type: `boolean | TypeHierarchyOptions | TypeHierarchyRegistrationOptions` where `TypeHierarchyOptions` is defined as follows:
+
+```typescript
+export interface TypeHierarchyOptions extends WorkDoneProgressOptions {
+	/**
+	 * The server has support for supporting inheritance tree.
```
puremourning, line 8225: "support for supporting" -> "support for providing an"

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+```typescript
+export interface TypeHierarchyPrepareParams extends TextDocumentPositionParams,
+	WorkDoneProgressParams {
+}
+```
+
+_Response_:
+
+* result: `TypeHierarchyItem[] | null` defined as follows:
```
puremourning, line 8253: Why should we allow `null` here ? What's the client expected to do with a null? It's better to define this to either return an empty array, or an error. Alternatively _explicitly_ define what the meaning of `null` is (as opposed to an empty array or an error return).

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+	/**
+	 * Indicates if this item has subtypes. When `undefined` the subtypes
+	 * of this item have not been resolved yet.
+	 */
+	hasSubtypes?: boolean;
+
+	/**
+	 * A data entry field that is preserved between a call hierarchy prepare and
+	 * incoming calls or outgoing calls requests.
```
puremourning, line 8309: "and incoming calls or outgoing calls requests" - I'm not sure I understand what that's saying.

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+```typescript
+export interface TypeHierarchySubtypesParams extends
+	WorkDoneProgressParams, PartialResultParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TypeHierarchyItem[] | null`
```
puremourning, line 8359: again, what's `null` for? we need to be explicit about what the valid responses are and what they mean, so that client and servers work together in harmony.

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+```typescript
+export interface TypeHierarchySupertypesParams extends
+	WorkDoneProgressParams, PartialResultParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TypeHierarchyItem[] | null`
```
puremourning, line 8336: what does `null` return mean?

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+```typescript
+export interface TypeHierarchyInheritanceTreeParams extends
+	WorkDoneProgressParams {
+	item: TypeHierarchyItem;
+}
+```
+_Response_:
+
+* result: `TreeItem<TypeHierarchyItem> | null`
```
puremourning, line 8393: What does `null` mean?

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,211 @@ export interface Moniker {
...
+
+#### <a href="#typeHierarchy_subtypes" name="typeHierarchy_subtypes" class="anchor">Type Hierarchy Subtypes(:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The request is sent from the client to the server to resolve the subtypes for a given type hierarchy item. The request doesn't define its own client and server capabilities. It is only issued if a server registers for the [`textDocument/prepareTypeHierarchy` request](#textDocument_prepareTypeHierarchy).
+
+_Request_:
+
+* method: 'typeHierarchy/subtypes'
```
puremourning, line 8348: It's not obvious why there are so many messages here, nor why the prepare and resolve steps are required.

For clients (and presumably many servers), it would be simpler to return the whole tree in a single message. With this current API proposal there will be a lot of chatter for an ostensibly simple dataset. Could we perhaps incorporate the ability for the server to return the full type hierarchy in response to just a simple `/typeHierarchy` message as the first instance? Do we know that this is extremely expensive to calculate in many servers (enough to justify the complexity of this back-and-forth prepare/request/request/request API?

It's also unclear to me when a client should use `subtypes` and `super types` requests vs the `inheritanceTree` request - unless I'm missing something, they are implementing the same thing in a different way? What's the rationale for this?

_specifications/specification-3-17.md
KamasamaK: It's copy/pasted from `CallHierarchyItem.data` and mistakenly unchanged.

_specifications/specification-3-17.md
CsCherrYY: Yeah, we don't need `boolean` here in this version. We are considering whether to remove the capability `inheritanceTreeSuppport` to keep the type hierarchy general. I'll update this PR then.

_specifications/specification-3-17.md
CsCherrYY: Yes. I'll fix it in the next commit.

_specifications/specification-3-17.md
CsCherrYY: Just as you mentioned, calculating the full type hierarchy is extremely expensive for a given type with many subtypes. We currently have an implementation in Java language server, it may cost about 25 seconds to calculate the **direct** subtypes of `java.io.Serializable` (1877 subtypes in the example project in total). We want the user to know what server is doing - after `prepare` request, the client can get the base type itself so that it can be shown. So the user can see the based type with some progress icons at least, with partial result mechanism (if implemented), the client can show the results gradually.

For the second question, we have two common views for all languages, they are `subtypes` and `supertypes`. For single inheritance languages, since a given class could only have one **superclass**, usually they have another view to show all the **classes** in the inheritance tree. That can be performed in many `supertypes` requests and a `subtypes` request as well, so we're still in discussing whether to isolate this request.

_specifications/specification-3-17.md
Eskibear: I find it was referring call hierarchy.
```ts
callHierarchyProvider?: boolean | CallHierarchyOptions | CallHierarchyRegistrationOptions;
```

which is equivalent to: 
``` ts
boolean // simplest 
OR
{
  workDoneProgress?: boolean;  // from WorkDoneProgressOptions
} 
OR
{
  workDoneProgress?: boolean;  // from WorkDoneProgressOptions
  id?: string;  // from StaticRegistrationOptions 
  documentSelector: DocumentSelector | null;  // from TextDocumentRegistrationOptions 
}
```

specifying whether server has capabitility to support work done progress / unregistration / customized scope of the feature.


> Is a simple boolean necessary/sufficient here since each server capability can be defined under TypeHierarchyOptions?

@KamasamaK  I think `boolean` is **insufficient** for advanced features mentioned above. For me, it's somehow **necessary** . Because on one hand, when I implement it in server side I might not care about any advanced feature, then I can simply return a `true`; on the other hand it's consistent with other capabilities like call hierarchy.

> We are considering whether to remove the capability inheritanceTreeSuppport to keep the type hierarchy general.

@CsCherrYY  I'm voting +1 for removing `inheritanceTreeSuppport`, otherwise for languages supporting multiple inheritance (e.g. C++), you may also need something like `inheritanceGraphSuppport`?

KamasamaK: If you don't care about implementing or can't implement advanced features, you can return `false` for their capabilities. For example, `codeLensProvider` only has a type of `CodeLensOptions` with one optional native capability, so that's already established for LSP. That is the reason I say it is probably unnecessary. But if the extra capability is going to be removed, that might change things.

Perhaps @dbaeumer can weigh in on the necessity of a simple `boolean` for new operations when *Options exists, and whether them having no or only optional native properties makes a difference in that determination.

Eskibear: Thank you for pointing out `CodeLensOptions`. It looks:
`{ resolveProvider: false }`  means server doesn't have a resolve provider.
`{ resolveProvider: true }`  means server also support to resolve code lens.
So now I have a question, if server doesn't support code lens at all, what should it return? (if boolean is not allowed)

_specifications/specification-3-17.md
CsCherrYY: In the `textDocument/prepareTypeHierarchy` request, if the element in the location of the params is not a valid type (that depends on the server implementation, whether to infer a valid type), it will return null to indicate that there is no result. For those types have no subtypes in `typeHierarchy/subtypes`, the server will return an empty array. 

Thanks for the comment, I'll add a description in the next commit.

CsCherrYY:
Thanks a lot for the comments from @KamasamaK @puremourning and @Eskibear , I have the following thinking and plan about updating this PR.
- Address the wording issues and mistakes you have mentioned.
- Remove `typeHierarchy/inheritanceTree` to make it general with both single inheritance languages and multiple inheritance languages, the support of inheritance tree could be implemented in clients.
- Remove `hasSupertypes` and `hasSubtypes` to keep the interface simple. The clients would keep the status of each item.
- As @puremourning said, it would be indeed simpler if we can return the full type hierarchy in one request, but performance issue is still crucial. Could we have a property naming like `depth` or `level` to indicate the resolve levels in the server? That allows the client to clear the resolve levels in one request. Since we're using partial result mechanism and the return type of `typeHierarchy/supertypes` or `typeHierarchy/subtypes` is an array, I have no promising idea to define a good data structure of the return array including more than one level of type hierarchy. Could we find a proper one?

 Do you have any suggestions or comments?

_specifications/specification-3-17.md
KamasamaK: Excluding `codeLensProvider` would indicate that the server does not support it.

> A missing property should be interpreted as an absence of the capability. If a missing property normally defines sub properties, all missing sub properties should be interpreted as an absence of the corresponding capability.

CsCherrYY:
Based on the current `typeHierarchy/supertypes` and `typeHierarchy/subtypes` requests, it's still useful for single inheritance languages(C#, Java, etc.) to support inheritance tree view, so we have a draft to support inheritance tree view without changing the current requests:
- in `TypeHierarchyOptions`, `inheritanceTreeSuppport` can be used to indicate if the language server has support for inheritance tree.

  ```
  export interface TypeHierarchyOptions extends WorkDoneProgressOptions {
      /** 
       * The server supports for providing an inheritance tree.
       */
      inheritanceTreeSuppport?: boolean;
  }
  ```
  Based on this, `boolean` could be removed from the candidates of the server capability property type. If the client knows the server supports inheritance tree view(`inheritanceTreeSuppport` is `true`), an inheritance tree view will be available.
- in `TypeHierarchySupertypesParams`, to distinguish between the general `typeHierarchy/supertypes` requests and the `typeHierarchy/supertypes` requests during the inheritance tree view initialization, `classOnly` is used to tell the server if the request should only return class:
  ```
  export interface TypeHierarchySupertypesParams extends
	  WorkDoneProgressParams, PartialResultParams {
	  item: TypeHierarchyItem;
	  classOnly?: boolean;
	  transactionId?: integer | string;
  }
  ```
-  In the inheritance tree view, the client will send multiple `typeHierarchy/supertypes` requests with its `classOnly` set to `true`, the request will return the super class of the given item. Based on this, the client could get all the ancestor classes of the base type in the inheritance tree. Another `typeHierarchy/subtypes` request will be sent to get the direct subtypes of the base type in the initialization stage. Among these requests, the same `transactionId` might help improve the performance in the server.


Does this make sense?

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,167 @@ export interface Moniker {
...
+#### <a href="#textDocument_prepareTypeHierarchy" name="textDocument_prepareTypeHierarchy" class="anchor">Prepare Type Hierarchy Request (:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The type hierarchy request is sent from the client to the server to return a type hierarchy for the language element of given text document positions. Will return `null` if the server couldn't infer a valid type from the position. The type hierarchy requests are executed in two steps:
+
+  1. first a type hierarchy item is prepared for the given text document position.
+  1. for a type hierarchy item the supertype or subtype type hierarchy items are resolved.
+
+In the first step, the `textDocument/prepareTypeHierarchy` request could have a unique, constant and optional `transactionId`. The following `typeHierarchy/supertypes` and `typeHierarchy/subtypes` requests in the second step could have the same `transactionId` in their params, which could be used to help indicate some cached data in the server.
```
jakebailey, line 8200: I'm not aware of any existing LSP API that uses this "prepare to get ID then pass ID back" method; this seems pretty specific to a particular LS implementation. The closest I'm aware of is completion (where it's arbitrary data), but that's because the user may not even visit a particular completion (so you want to defer that work to later in hopes it's never done). This new API appears to always call prepare, then one of the two other calls.

Why not just have the two calls and cache the information internally (say, per snapshot or similar?). This would match other similarly-expensive calls like go-to-references or go-to-implementations which are simple requests. If we were to add this to Pylance, we wouldn't need this transaction ID, as the info is there in the analysis (or will be lazily produced from the current state)

I can also see this pattern being problematic for LSIF (where you expect basically pure functions for read-only requests), but I guess the ID is optional.

Eskibear: > this "prepare to get ID then pass ID back" method

As far as I understand of this proposal, client is reponsible to generate this ID itself, and passes it to server in request `textDocument/prepareTypeHierarchy`. I assume the purpose of this "prepare" request is to get corresponding `TypeHierarchyItem` of the cursor position where you send the request, instead of getting ID.

As mentioned above in https://github.com/microsoft/language-server-protocol/pull/1231/files#r612267620 (though not explicitly described in spec) if it's not a valid type (e.g. triggered on some keywords), response of "prepare" request can be null, and client probably should not send further requests then.

jakebailey: I can sort of see the argument of needing to re-query at various points in the same hierarchy, but I can see troubles where because the client is defining this ID, it doesn't have enough information to determine when the result has actually changed (e.g., the server has changed the analysis and now this request cycle is invalid; like on an external file watcher change), and therefore requires something like how semantic tokenization has notifications for "the data has changed".

One thing that I think is similar to this right now is signature help retriggering; as the user moves the cursor, the previous signature info is passed back to the server, which can then use that to further edit the response, or, choose to ignore it and give the data fresh.

I guess I'll need to reread this with fresh eyes in the morning.

lastthyme (approved):

lastthyme (approved):

lastthyme (approved):

lastthyme (approved):

lastthyme (approved):

_specifications/specification-3-17.md
dbaeumer: Here is an alternative approach: the prepare request returns a TypeHierarchyItem so the data field of the item can be used to identify the type hierarchy. If the type hierarchy changes I would suggest that the server errors this on the next sub or super type request and lets the client decided what to do (e.g. refresh the hierarchy, close the hierarchy, ...)

_specifications/specification-3-17.md
dbaeumer: `null` is for consistency since lots of LSP requests that return and array can return `null` to indicate no elements

_specifications/specification-3-17.md
jakebailey: Yeah, I find that to be a bit more understandable and close to completion resolution / signature help callbacks (assuming this new data field is something returned by the server at the prepare call, then passed back by the client, versus how this spec currently has the client come up with an ID).

CsCherrYY: > Here is an alternative approach: the prepare request returns a TypeHierarchyItem so the data field of the item can be used to identify the type hierarchy. 

So under this approach, in my understanding, the server will be responsible for managing the "transaction" itself, and client always keep it unchanged in the `TypeHierarchyItem` and send it back with the next sub or super type requests, is that right?

And if we apply this and remove `transactionId` in params, should we add some description explicitly to show that the `data` field could be used to carry the type hierarchy related information in the server? I think it may help the designers of language servers if the server can use this mechanism to improve the performance.

dbaeumer: @CsCherrYY yes, that is the idea and consistent with other situations. For example the completion item has a data field as well which the clients needs to preserve and sent back on a resolve call so that the server can related the completion item to some previous request. 

A comment in the data field that describes the use case of a transaction ID is for sure something we should do.

CsCherrYY:
I have updated the PR based on @dbaeumer 's [comment](https://github.com/microsoft/language-server-protocol/pull/1231#discussion_r622242278) and remove `transactionId`. Besides, the `inheritanceTreeSuppport` is added as a server capability in this commit as well, the draft is mentioned in https://github.com/microsoft/language-server-protocol/pull/1231#issuecomment-823074083.

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,172 @@ export interface Moniker {
...
+	 */
+	selectionRange: Range;
+
+	/**
+	 * A data entry field that is preserved between a type hierarchy prepare and
+	 * supertypes or subtypes requests. It could also be used to identify the 
+	 * type hierarchy in the server, helping improve the performance on 
+	 * resolving supertypes and subtypes.
+	 */
+	data?: unknown;
```
jakebailey, line 8301: Would it make more sense for the prepare response to be an object (versus an array), and have a single data field? I guess I'm not sure to what extent the full call needs data versus each individual item (where if you really just wanted the same data, you'd have to dupe it on each of them).

CsCherrYY: I think if we need a data field in each individual item would depend on server. If server could hold or cache the whole type hierarchy and use it to find a type in coming requests, one data field is OK, but if server wants to use `data` field to indicate the specific type, we'd better to keep a data field in each item. In protocol, I suggest keeping this field in each item to support these two usages both.

_specifications/specification-3-17.md
```diff
@@ -8173,6 +8188,172 @@ export interface Moniker {
...
+
+* property name (optional): `typeHierarchyProvider`
+* property type: `TypeHierarchyOptions | TypeHierarchyRegistrationOptions` where `TypeHierarchyOptions` is defined as follows:
+
+```typescript
+export interface TypeHierarchyOptions extends WorkDoneProgressOptions {
+	/** 
+	 * The server supports for providing an inheritance tree.
+	 */
+	inheritanceTreeSuppport?: boolean;
```
Eskibear, line 8227: typo in support

dbaeumer: I am not convinced that we should add `inheritanceTreeSuppport`  in the first version since I don't see clients to support this rendering. IMO it makes things simply harder to understand. Or do I miss something?

CsCherrYY: My apologize for not mentioning the client behaviors about `inheritanceTreeSupport` in the protocol. I'd like to describe it in detail here:
 
When using `inheritanceTreeSupport`, the client has the same mechanism of handling response data as the other two requests. The only difference is how to set `classOnly` in `TypeHierarchySupertypesParams`. 
- If the server doesn't set `inheritanceTreeSupport` or it's `false`, the `classOnly` will always be `undefined` since the server doesn't support inheritance tree. Besides, **Inheritance Tree** button will hide in the client.
- When the server set `inheritanceTreeSupport` to `true`, the client could show **Inheritance Tree** button.
  - If the active view is inheritance tree view, after prepare request, to get the most ancestor class of the given type, the client will send several `typeHierarchy/supertypes` requests with `classOnly` in their params set to `true`. The responses include all classes in the inheritance tree of the given item, so that the client can render this tree reversally. The remaining requests work normally.
  - If the active view is not inheritance tree view, `classOnly` will always be `false` and everything works normally.
- In the server, if there is a `typeHierarchy/supertypes` request with its `classOnly` is `true`, the response should contain class only, otherwise it contains all types.

If there is no `inheritanceTreeSupport` in the protocol, since we'd like to support inheritance tree view for it's the most important type hierarchy view for Java users (might also for single inheritance language users), we have to independently implement the view in java extension instead. We might make it as a command and use `workspace.executeCommand` to trigger this view. In server, we'll have the similar logics to handle the supertype requests which return only class. Another concern is that if we can put the **Inheritance Tree** button properly. See the following screenshot:
type2
Currently we could put them together since we contribute `ms-vscode.references-view`, if the right two buttons are supported in LSP and implemented in client already, can we still find a way to put them together?

What do you think of this?

dbaeumer: I was incorrect about what I said. I see the rendering but I am still not convinced for the modes in the API. Explaining this is simply too complicated. Why can a client not render that tree using the existing API by filtering interfaces. And what do we expect that happens if a client calls the API with `classOnly` for a item that represents a interface.

CsCherrYY: > Why can a client not render that tree using the existing API by filtering interfaces. 

A client can render that tree by filtering interfaces indeed. It's a good tradeoff to remove `classOnly` and put all the logics in client implementation to keep the protocol clear. If we remove `classOnly` and keep `inheritanceTreeSupport` in the first version, should we define the behaviors of client? If we remove `inheritanceTreeSupport` from the first version, my concern is whether we can still keep the **inheritance tree view** (from java extension) and the other two views button (from VSCode) together. 
> And what do we expect that happens if a client calls the API with `classOnly` for a item that represents a interface.

Currently we expect calling the API with `classOnly` for an interface will return an empty array since never a class can be the supertype of an interface.

HighCommander4: One thing that may be worth keeping in mind is that some languages like C++ have multiple class inheritance, so `classOnly` may not imply only one supertype.

dbaeumer: I am not understanding why `inheritanceTreeSupport` can influence how a client can renderer a class hierarchy. Isn't this something that is under the sole control of the client and a server should never influence this?

Another problem I see with `classOnly` is that it is specific for a certain use case. It might cause problems with other languages. I could for example imagine a special view for language that have multiple class inheritance. Another view could be special for interface to see the implemented protocols. 

If we think that that performance and payload size is a problem then we should add a filter property to the params which would allows to filter hierarchy items on the server side. IMO `classOnly` is simply to specific for a specific use case.

Eskibear: #### About inheritanceTreeSupport
After reading through above comments, I think current naming is causing confusion. Let me help to clarify according to my understanding: 

`ServerCapabilities.inheritanceTreeSupport` proposed by @CsCherrYY indicates whether _"the language **only** supports single inheritance through **class**"_. E.g. it's true for Java because multiple inheritance is supported through interfaces. And it's false for C++.  (@CsCherrYY correct me if I'm wrong)

"how a client can renderer a class hierarchy" mentioned by @dbaeumer  should be related to a **Client** capability if needed. For example `inheritanceTreeSupport`, `inheritanceGraphSupport` (for multiple-inheritance languages?)  etc. (@dbaeumer correct me if I'm wrong)



#### About classOnly
> If we think that that performance and payload size is a problem then we should add a filter property to the params

Agree that `classOnly` is merely a specific filter property, is there any other filtering requirement (e.g. interfaceOnly), how would a generic filter property look like?
I'm testing this proposal on Java language server, so far I don't see any performance/payload size problems as we expand the hierarchy level by level, and I'm fine to filter items on either server/client side. 

Here I only have a stupid question as I know little about client implementation details: No matter what the filtering property eventually is, if I want to support showing class view (e.g. using tree items in vscode's reference view) for Java, how does client know when to filter the items or not? Does it provide multiple entries (e.g."show supertypes", "show subtypes", "show class view", "show interface only") and send requests with different params corresponding to the entries?

Eskibear:
I tried implementing current "prepare/supertypes/subtypes" proposal for Java on VS Code. Below is a demo based on type `B1`.

image


https://user-images.githubusercontent.com/2351748/120151402-ca134500-c21e-11eb-9b63-73cfcd17f169.mp4

Eskibear:
About performance and payload, I tested with interface `java.io.Serializable` which has ~1800 subtypes, it took 5-7s for the subtypes request. And it took ~60s to fetch subtypes of `java.lang.Object`, i.e. all types including anonymous ones.

https://user-images.githubusercontent.com/2351748/120162592-211f1700-c22b-11eb-9756-eaf83bd86f29.mp4


Above was tested on Intel Core i7-10700. Server implementation depends on how powerful the CPU is, e.g. the `Serializable` case took ~20s for another machine with Intel Xeon E5-1620 v4 (less powerful).

rcjsuen (commented):
For `typeHierarchy/supertypes` and `typeHierarchy/subtypes`, should the request parameters use something simpler than a `TypeHierarchyItem`. Is it necessary to include `range` and `selectionRange` when asking the server to fetch this information?

CsCherrYY:
> For `typeHierarchy/supertypes` and `typeHierarchy/subtypes`, should the request parameters use something simpler than a `TypeHierarchyItem`. Is it necessary to include `range` and `selectionRange` when asking the server to fetch this information?

`range` and `selectionRange` could be used to specify the particular type in the server. In some cases, like resolving anonymous types in Java, we can't use only `name`, `uri` and `detail` to find the exact type. (Another way is to use `data` field to preserve some identifier, but `data` is not a required field)

_specifications/specification-3-17.md
dbaeumer: Filter properties would be properties that are available on a TypeHierarchyItem. I would say typically the  kind property.

Regarding the UI: it could either be a different command or some sort of filter buttons in the UI.

dbaeumer:
@rcjsuen I was thinking the same but IMO this doesn't cause a lot of overhead and keeps the spec simpler. We don't need to define another type.

Regarding anonymous types: I think this are best identified using the data property on an item.

@CsCherrYY the performance is mainly influenced by the server producing the data. Not by sending them to the client. Right?

CsCherrYY:
> @CsCherrYY the performance is mainly influenced by the server producing the data. Not by sending them to the client. Right?

Yes. The time for producing the data in the server takes up major time. I just tried @Eskibear 's implementation on my PC. 

For `java.io.Serializable` (1877 subtypes), the total time to receive the response of `java/subtypes` is **23172ms**, the time when producing the data in the server is **23082ms**, which takes up about **99.6%** of total time.

For `java.util.List` (37 subtypes), the total time to receive the response of `java/subtypes` is **2191ms**, the time when producing the data in the server is **2176ms**, which takes up about **99.3%** of total time.

dbaeumer:
@CsCherrYY thanks for clarification.

_specifications/specification-3-17.md
CsCherrYY: If it's possible to keep the Class Hierarchy view available for the Java users, I'm very glad to remove `inheritanceTreeSupport` in the first version since it's indeed a little complicated. Here is another approach to support Class Hierarchy: The Java extension can contribute another button in reference-view and use a separated command to show class hierarchy.  But there is still a concern about the UI implementation.

We want to make the separated class hierarchy button align with the two existing type hierarchy buttons (supertype hierarchy and subtype hierarchy). In the current reference-view implementation, the call hierarchy uses [context value](https://github.com/microsoft/vscode-references-view/blob/870c7f3098f10090581215fbc7fcf1af23e19b40/package.json#L199) to control whether to show the button. Since the call hierarchy has only two views, it works well. But for type hierarchy, can we make the context value extendable to support possible other views? Then we can use other conditions like `enablement` to show if this button is disabled, like the current [Java implementation](https://user-images.githubusercontent.com/45906942/118770745-466b6700-b8b4-11eb-8e20-fcad056611d6.jpg).

@dbaeumer Could you give any suggestions?

CsCherrYY: Update: Another concern is about the default view. Currently the default type hierarchy view for java users is Class Hierarchy view, which is most frequently used by java developers. However, the implementation of type hierarchy in [vscode-references-view](https://github.com/microsoft/vscode-references-view) would be based on the protocol, supporting subtype hierarchy view and supertype hierarchy view only. Could it be possible to change the default view of type hierarchy in another extension? e.g. Java extension.

dbaeumer:
@CsCherrYY as a next step we should add an implementation of the type hierarchy to the LSP client libraries for VS Code.

CsCherrYY:
> @CsCherrYY as a next step we should add an implementation of the type hierarchy to the LSP client libraries for VS Code.

Sure. If we implement this protocol without `inheritanceTreeSupport`, there are still some concerns about the implementation, as I posted previously in https://github.com/microsoft/language-server-protocol/pull/1231#discussion_r647414065 and https://github.com/microsoft/language-server-protocol/pull/1231#discussion_r647989112:
- For the UI: We would like to find a proper way to make the class hierarchy view button (will be contributed by Java extension) align with the supertype view button and subtype view button.
- For the default view: We would like to make it possible to change the default view if the user is currently using Java extension.

@Eskibear is working on corresponding implementation for LSP client libraries already. I think we can focus on the implementation once we have resolved these concerns.

dbaeumer:
@CsCherrYY regarding  #1231 (comment) 

To my understanding this is a problem with the reference view. It has nothing to do with what we define in the protocol. Right? 

If this is true then we need to discuss the button problem with the VS Code team and might need to think about a PR against https://github.com/Microsoft/vscode-references-view.

CsCherrYY:
> @CsCherrYY regarding #1231 (comment)
> 
> To my understanding this is a problem with the reference view. It has nothing to do with what we define in the protocol. Right?
> 
> If this is true then we need to discuss the button problem with the VS Code team and might need to think about a PR against https://github.com/Microsoft/vscode-references-view.

Yes, they are UI related problems and I'll create an issue to talk about this in the [reference-view](https://github.com/Microsoft/vscode-references-view) repository.

CsCherrYY:
@dbaeumer  I have remove the part about `inheritanceTreeSupport` in https://github.com/microsoft/language-server-protocol/pull/1231/commits/90090af89be759fe116821b707a0ce53bc214bea and create https://github.com/microsoft/vscode/issues/126666, we can discuss the UI problems there.

isc-bsaviano:
What is the status of this PR? The TypeHierarchy API in VS Code is getting finalized for the upcoming release and the PR implementing this in `vscode-languageserver-node` was merged a while ago

dbaeumer:
We will have a new propsed version of the libs beginning of October. Depending on the feedback there might be changes. When we have some confidence that all is good I will copy the MD part from the implementation to the spec. I would like to keep the PR open to not forget this.

KamasamaK:
Type Hierarchy has already been [finalized for VS Code](https://github.com/microsoft/vscode/commit/cbf8b411b9ccce33926bfa2d20f86e0970a85235). Is anything still pending for this PR?

dbaeumer:
I wanted to wait with merging the spec to see if there is some feedback about the proposed implementation which is available since a while in the VS Code libs.

KamasamaK:
This was added to the spec in 7201ba1db625708daf3d81bcd0e6433a923e5830

dbaeumer:
Correct. Will close the PR

PR #1245, opened 2021-04-15
Add CodeActionTriggerKind

### LINKED ISSUE #1244: CodeActionContext.triggerKind
From latest VSCode notes: https://code.visualstudio.com/updates/v1_55#_codeactioncontexttriggerkind

I assume this is intended for eventual inclusion in LSP (and needs to become stable in VSCode first), but wanted an issue to track.

This would be really useful for clangd: we've struggled a lot with the signal/noise ratio of code actions and conflation of available vs suggested actions. (Both in VSCode & other editors).

KamasamaK:
Implements the newly finalized API from VS Code for the `CodeActionTriggerKind` in `CodeActionContext`.

Fixes #1244

### DIFF
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -6152,6 +6152,27 @@ export interface CodeActionParams extends WorkDoneProgressParams,
 }
 
 /**
+ * The reason why code actions were requested.
+ * 
+ * @since 3.17.0
+ */
+export namespace CodeActionTriggerKind {
+	/**
+	 * Code actions were explicitly requested by the user or by an extension.
+	 */
+	export const Invoked: 1 = 1;
+
+	/**
+	 * Code actions were requested automatically.
+	 *
+	 * This typically happens when current selection in a file changes, but can
+	 * also be triggered when file content changes.
+	 */
+	export const Automatic: 2 = 2;
+}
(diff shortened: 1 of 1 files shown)


KamasamaK:
I didn't see a need for a client capability since client support would not affect what the server sends.

_specifications/specification-3-17.md
```diff
@@ -6241,6 +6262,13 @@ export namespace CodeActionKind {
  * a code action is run.
  */
 export interface CodeActionContext {
+	/**
+	 * The reason why code actions were requested.
```
puremourning, line 6266: should we define what the absence of this value means. e.g. that the server should make no assumptions about the kind of trigger and return all code actions that are relevant ?

dbaeumer: Makes sense. But it is also somehow implicit since we didn't have that value before and servers needed to return all values.

KamasamaK:
Closing since commit 27c15c694a80152f918b2b60dbf7ff09f1bb909e was already added by @dbaeumer. The only difference I see is the div anchor, which I don't think was used at the time I created this.

dbaeumer:
@KamasamaK Ihave to apologies. I saw the issue yesterday and thought I can quickly address it without looking if there is a pull request. I also implemented it in the libs.

PR #1250, opened 2021-04-20
proposal: documentation extension to LSIF

slimsag:
This is a proposal to add a documentation extension to LSIF.

The aim of this extension is to enable LSIF indexers to produce API documentation
that is competitive with the API documentation offered in some languages today
through various websites such as:

* (Go) https://pkg.go.dev
* (Rust) https://docs.rs
* (Java) https://javadoc.io
* (Zig) https://ziglang.org/documentation/master/std/#builtin

This will be an LSIF-only extension, not an extension to LSP. The motivation for
doing this is outlined in https://github.com/slimsag/language-server-protocol/pull/1

A Go implementation of this same spec [is available here](https://github.com/sourcegraph/sourcegraph/pull/20108).

Helps sourcegraph/sourcegraph#19389

Signed-off-by: Stephen Gutekanst <stephen@sourcegraph.com>

### DIFF
--- a/indexFormat/specification.md
+++ b/indexFormat/specification.md
@@ -1044,6 +1044,131 @@ export interface MetaData {
 }
 ```
 
+## Sourcegraph Extensions
+
+### The Documentation vertex
+
+A Documentation vertex describes hierarchial project-wide documentation.
+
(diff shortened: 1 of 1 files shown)


slimsag:
Wrong repository, sorry :) this is still early stages & not yet ready to propose here

PR #1251, opened 2021-04-22
'message' to 'messages' in TraceValue

giltho:
I introduced that mismatch when specifying the behaviour of setTrace apparently, see [this diff].(https://github.com/microsoft/language-server-protocol/pull/953/files)
'messages' is still the one being used by VSCode, and it has been an issue for [ocaml-lsp](https://github.com/ocaml/ocaml-lsp/pull/425)

the standard implementation still uses 'messages' as it was before that change


dbaeumer:
Thanks for the correction!

PR #1253, opened 2021-04-22
Add ltex-ls to list of language servers

valentjn:
I'd like to add the LanguageTool language server [ltex-ls](https://github.com/valentjn/ltex-ls) (maintained by me) to the list of language servers on the website. It's a fork of Adam Voss's language server, which is already in the list but which is unmaintained since 2017 due to the passing of its author. It should be significant enough as there are language clients for VS Code (&ldquo;reference client&rdquo; [vscode-ltex](https://github.com/valentjn/vscode-ltex), also maintained by me), Sublime Text, Emacs, and Neovim. Server and reference client have a pretty extensive [documentation](https://valentjn.github.io/vscode-ltex/docs/settings.html) of the configuration.


valentjn:
Is there anything left to do for me in order to merge this?

PR #1257, opened 2021-05-01
Remove Content-Type header

### LINKED ISSUE #1256: Deprecate Content-Type header
Removing `Content-Type` from the spec will take less burden on implementations (#600), because there are no known uses for it, and defaults are assumed by spec.

abitrolly:
Fixes #1256

### DIFF
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -29,17 +29,14 @@ separated by a '\r\n'.
 
 ### <a href="#headerPart" name="headerPart" class="anchor"> Header Part </a>
 
-The header part consists of header fields. Each header field is comprised of a name and a value, separated by ': ' (a colon and a space). The structure of header fields conform to the [HTTP semantic](https://tools.ietf.org/html/rfc7230#section-3.2). Each header field is terminated by '\r\n'. Considering the last header field and the overall header itself are each terminated with '\r\n', and that at least one header is mandatory, this means that two '\r\n' sequences always immediately precede the content part of a message.
-
-Currently the following header fields are supported:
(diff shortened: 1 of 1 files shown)


lastthyme (approved):

puremourning:
Won’t This make all existing implementations non conforming. Hard to see that as a good thing?

abitrolly:
The header is optional. Protocol doesn't specify what should server do with unknown headers.

puremourning:
By removing it from the spec, you make it no longer "optional" but "forbidden" which I think makes all existing implementations that send it suddenly non-conforming.

abitrolly:
@puremourning this makes implementations that for decided to send this header for some reason (which?) not supporting latest spec. Which happens with every new spec version, no? It would be nice to answer in spec what implementations are sending this header and what for. If there is no reason for them to send it, and the header is useless, then if they can send it "just in case" is the protocol design mistake that this PR is trying to fix.

puremourning:
> Which happens with every new spec version, no? 

Not as far as I can tell. So far every change has been backwardly compatible. The spec versions are just for reference; the protocol is not versioned in the sense of sending a version header. Compatibility is maintained by not making breaking changes.

Regarding which clients or servers use this header, then I guess if proposing to change it, there's a real burden on the proposer to prove that it wouldn't cause actual regression. I'm just pointing out that at leat in a _technical_ sense this change makes previously-conforming implementations non-conforming which I think is an error, unless we can prove that it isn't.

abitrolly:
> So far every change has been backwardly compatible. 

There could be a clause that server should ignore unknown headers. This will also make transition to REST API easier if JSON-RPC will appear to be limited/slow. Sounds good?

abitrolly:
(off-topic comment regarding JSON-RPC - I would have saved a lot more time if I could use plain `curl` for developing and debugging server problems)

dbaeumer:
See my comment in https://github.com/microsoft/language-server-protocol/issues/1256

abitrolly:
I disagree with your comment. You provided an opinion, but no explanation.

PR #1282, opened 2021-06-02
Clarify isIncomplete feature behavior.

NTaylorMullen:
- When playing around with isIncomplete it was unclear from the spec perspective on the appropriate handling of isIncomplete. Clarifying that isIncomplete should represent "replaced" items and not "appended" items makes it clear how a language server should handle the feature.
- Ported the wording to all previous specs + fixed some typos.


dbaeumer:
@NTaylorMullen thanks!

PR #1292, opened 2021-06-10
Add Terraform & HCL language IDs

radeksimko:
The commit message and modification should be self-explanatory, but let me know if there's anything I missed about how this list gets updated!


dbaeumer:
Added Elixir and Erlang.

PR #1293, opened 2021-06-10
Add ccls to the list

NickStrupat:


i-ky:
As far as I know, ccls is not an SDK, but a language server. And it has been listed among [language servers](https://microsoft.github.io/language-server-protocol/implementors/servers/) long before this PR (see #525):
https://github.com/microsoft/language-server-protocol/blob/50732421db206ec7dd5c30b88326083bcaf826ef/_implementors/servers.md?plain=1#L37

NickStrupat:
You're right, I was confused. Sorry!

PR #1296, opened 2021-06-13
Fix: missing empty lines between html end tags and start of code blocks

rwols:
See: https://spec.commonmark.org/0.29/#example-131

### DIFF
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -70,6 +70,7 @@ The following TypeScript definitions describe the base [JSON-RPC protocol](http:
 The protocol use the following definitions for integers, unsigned integers and decimal numbers:
 
 <div class="anchorHolder"><a href="#integer" name="integer" class="linkableAnchor"></a></div>
+
 ```typescript
 /**
  * Defines an integer number in the range of -2^31 to 2^31 - 1.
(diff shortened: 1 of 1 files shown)


rwols:
I'm starting to think it's intentional to not have an empty line 6f0790fad6b7

rwols:
... but example 155 (https://spec.commonmark.org/0.29/#example-155) states that this input

````
<div>
bar
</div>
*foo*
````

is processed to

````
<div>
bar
</div>
*foo*
````

which seems to suggest that, for portability, an empty line is actually required

KamasamaK:
I don't think it's intentional. Their newer PR #1287 explains that those newlines should be there.

NTaylorMullen:
So technically newlines aren't required functionally from my experiments; however, I found that without them you lose VSCode colorization of the TypeScript blocks. Therefore, to ensure a nice editing experience we should have newlines after all of them 😄

NTaylorMullen (approved):

rwols:
Did my commit vanish into thin air? :D I don't see it on the gh-pages branch.

NTaylorMullen:
To be honest I thought I had addressed your fixes a lot earlier and then after seeing your PR figured I must have missed a few. It's possible your gh-pages wasn't sync'd when you initially made the changes :)

rwols:
Alright, sounds good to me

PR #1308, opened 2021-07-01
Add source.fixAll constant

DanTup:
@dbaeumer I don't know if this is reasonable, but VS Code has a constant for it and it seems like it would be useful for the same reason as having `source.organizeImports` here - so clients can provide better UI beyond just the Source Actions list for invoking this kind of (common) operation.

### DIFF
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -6684,6 +6684,16 @@ export namespace CodeActionKind {
 
 	/**
 	 * Base kind for an organize imports source action:
+	 * `source.fixAll`.
+	 *
+	 * Fix all actions automatically fix errors that have a clear fix that
+	 * do not require user input. They should not suppress errors or perform
+	 * unsafe fixes such as generating new types or classes.
+	 */
+	export const SourceFixAll: CodeActionKind = 'source.fixAll';
+
+	/**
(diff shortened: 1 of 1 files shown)


dbaeumer:
Makes sense.

_specifications/specification-3-17.md
```diff
@@ -6682,6 +6682,16 @@ export namespace CodeActionKind {
 	 */
 	export const Source: CodeActionKind = 'source';
 
+	/**
+	 * Base kind for an organize imports source action:
+	 * `source.fixAll`.
+	 *
+	 * Fix all actions automatically fix errors that have a clear fix that
+	 * do not require user input. They should not suppress errors or perform
+	 * unsafe fixes such as generating new types or classes.
```
rchl, line 6691: Some copy-paste and wording issues:

This is wrong:

> Base kind for an organize imports source action:

This sentence doesn't start well:

> Fix all actions automatically fix errors

DanTup: Eugh, good catch. The second part was copy/pasted from VS Code so I didn't want to change it too much. I've added quotes around 'Fix all' that I think make it clearer - I've filed https://github.com/microsoft/language-server-protocol/pull/1309 - comments welcome. Thanks!

PR #1310, opened 2021-07-02
Fix a typo.

teymour-aldridge:


KamasamaK:
Thanks for this fix! Can you please also apply this fix for `specification-3-17.md`?

PR #1315, opened 2021-07-10
Update comment for jsonrpcReservedErrorRangeEnd

leegbestand:
The comment states that it is the start of the range, but the variable name states it is the end. The variable name seems to be correct, so update the comment.

### DIFF
--- a/_specifications/specification-3-16.md
+++ b/_specifications/specification-3-16.md
@@ -202,7 +202,7 @@ export namespace ErrorCodes {
 	export const UnknownErrorCode: integer = -32001;
 
 	/**
-	 * This is the start range of JSON RPC reserved error codes.
+	 * This is the end range of JSON RPC reserved error codes.
 	 * It doesn't denote a real error code.
 	 *
 	 * @since 3.16.0


rcjsuen:
```TypeScript
export const jsonrpcReservedErrorRangeStart: integer = -32099;
```
```TypeScript
export const jsonrpcReservedErrorRangeEnd = -32000;
```
Looks like we have a bigger problem where the "end" is `-32000` and the "start" is `-32099`...?

KamasamaK:
@rcjsuen The `lspReservedErrorRangeStart` and `lspReservedErrorRangeEnd` codes are the same, so at least it's consistent. It's not not exactly intuitive, but it makes sense when you think about the start of a range most commonly being the lesser value.

The JSON-RPC 2.0 spec isn't even consistent on this for their [Error object](https://www.jsonrpc.org/specification#error_object). In the text, it says "The error codes from and including -32768 to -32000 are reserved for pre-defined errors" for their entire reserved range, but in the table it's "-32000 to -32099" for the reserved server error range.

dbaeumer:
I named them that way since you usually get a next error by adding +1. Constantly doing this brings you from start to end.

PR #1323, opened 2021-07-23
Adjusted alphabetical sorting by language

vintagedave:
The first column of the language server table is the Language, and it is mostly alphabetically sorted. A year or more ago, last time I checked this list, it was fully alphabetically sorted by the language name. However today there are a few entries that were not placed in the table following that precedent, and this makes it confusing to read or scan the table. This moves those confusing items so the table is once again alphabetically sorted by language.

I have moved:
* Deno to between Delphi and Dockerfiles
* The IBM servers for High Level Assembler to A for Assembler (see note below)
* The IBM servers for COBOL to C, for COBOL
* Crane PHP to PHP
* JSON to the end of the J section
* T-SQL from between R and S, to T. This is because T-SQL is not quite the same as SQL [according to Wikipedia](https://en.wikipedia.org/wiki/Transact-SQL) (note: it was not sorted as SQL anyway. Note 2: should it be called Transact-SQL?)

Of course, a request to change how something is sorted by someone more familiar with the language (I know only what Google shows about T-SQL vs SQL, for example) is very welcome. The goal here is to clean up the table.

Where the language is now not the first word in the table cell, I have bolded the language name: "IBM Enterprise **COBOL** for z/OS". Scanning the table, this seems to read well and makes the language name clear in the language column. As well as the ones I moved, I added this formatting to IBM PL/I as well which was already sorted as P for PL/I, same for IBM LALR (already under L.)

Most languages here just list the language, not the product, whereas other than Deno the confusingly sorted ones mix in a product or brand name with the language (eg Crane PHP is, as far as I can tell, a language server for PHP, not a fork of the PHP language itself; IBM COBOL is, well, COBOL; the IBM assemblers are more difficult since assemblers are different for different targets, and these are for z/OS, but I sorted under A for Assembler. IMO this is fine since there are no servers listed for other assembly languages yet.) Another alternative would be to remove all product or branding in the Language column, and either add a new column or use the Maintainer column for that information, keeping the language column only for the language name.

### DIFF
--- a/_implementors/servers.md
+++ b/_implementors/servers.md
@@ -19,6 +19,8 @@ index: 1
 | [API Elements](http://api-elements.readthedocs.io/en/latest/) | [Vincenzo Chianese(@XVincentX)](https://github.com/XVincentX) | [vscode-apielements](https://github.com/XVincentX/vscode-apielements) | TypeScript |
 | Apache Camel | [Contributors](https://github.com/camel-tooling/camel-language-server/graphs/contributors) | [Apache Camel Language Server](https://github.com/camel-tooling/camel-language-server) | Java |
 | Apex | Salesforce |  [VS Code Apex extension](https://marketplace.visualstudio.com/items?itemName=salesforce.salesforcedx-vscode-apex) | TypeScript |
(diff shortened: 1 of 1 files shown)


dbaeumer:
Thanks!

PR #1329, opened 2021-08-04
http://tools.ietf.org/html/rfc3986 is no longer a valid URL

fredg1:
tools.ietf.org seems to have been sold to someone else since 2006. 2006!!!


davidhalter:
It seems to work only with https, try this link instead:

https://tools.ietf.org/html/rfc3986

dbaeumer:
Change to `https`

PR #1334, opened 2021-08-17
Sort list of supporting tools alphabetically

fingolfin:
That makes it arguably easier to scan the table visually.

I also changed the two entries labelled `emacs` resp. `GNU Emacs` to both say `Emacs`
(an alternative would be to leave them as `GNU Emacs` but sort them under `Emacs`).

There are some more entries one may argue about, e.g. should "GNOME Builder" be sorted
under "G" or "B"? What about "MS Monaco"?

Finally, one entry says "Multiple editors"; the underlying ycmd tool supports Emacs,
vim, atom, sublime, geany, kakoune, nano, ... perhaps it should be listed multiple times?

### DIFF
--- a/_implementors/tools.md
+++ b/_implementors/tools.md
@@ -10,41 +10,41 @@ index: 2
 
 | Editor/IDE | Maintainer | Repository |
 |------|--------|----------|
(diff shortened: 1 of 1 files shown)


dbaeumer:
Thanks

PR #1342, opened 2021-09-06
fix links to code types in specification 3.14, 3.15 and 3.16

corentinguillerme:
Currently, all links to code types lead to 3.17 page, even from other versions


dbaeumer:
@NTaylorMullen can you please have a look.

The idea of the linking is still that we can restructure the document. To not force use to do this for older versions I would argue that we better not offer linking for the old versions. I want to ensure we are not causing any maintenance effort in those. 

@corentinguillerme I see that you did the work for the old documents. It might not make it due to the argument above. I hope that is OK for you.

corentinguillerme:
I see, it's OK for me. 

In that case, maybe it would be better to remove the links in the old documents instead of linking them to another version (as it is now)?
This will prevent a user from unintentionally changing versions by clicking on these links.

dbaeumer:
@NTaylorMullen friendly ping

NTaylorMullen:
> To not force use to do this for older versions I would argue that we better not offer linking for the old versions.

Ya that's a fair point. However, we couldddd potentially not change the older document styles/structure and only change the newer ones. Not sure how much we care about consistency

dbaeumer:
I think it is fine to only update the 3.17 specification with the links.

dbaeumer:
@NTaylorMullen could you please look into scoping the link work to >= 3.17.

NTaylorMullen:
https://github.com/microsoft/language-server-protocol/pull/1373

dbaeumer:
@NTaylorMullen  #1373 didn't work. See my comment in the PR.

dbaeumer:
I will close the PR since it got very stale due to inactivity.

PR #1356, opened 2021-10-04
fix a11y issues

sana-ajani:
same issues as on DAP site: https://github.com/microsoft/debug-adapter-protocol/pull/217


dbaeumer:
Thanks a lot!

PR #1373, opened 2021-11-05
Restrict linking to only the 3-17 specification.

NTaylorMullen:
- I tried making linking >= 3-17; however, that wasn't easily doable due to the structure of the jekyll site. The JSFiles themselves are included in the layout which don't have access to page level information. Because of this I've hacked a hardcoded "if 3.17" check into the injected JavaScript to support linking. Once we publish 3.17 we'll need to go back and update the logic to ensure that 3.17+ is supported.

Originally attempted fix here: https://github.com/microsoft/language-server-protocol/pull/1342

### DIFF
--- a/_includes/js_files.html
+++ b/_includes/js_files.html
@@ -17,7 +17,6 @@
 <script async defer src="https://buttons.github.io/buttons.js"></script>
 <script src="{{ site.baseurl }}/js/page.js"></script>
 
-
 <script type="text/javascript">
   var linkableTypes = {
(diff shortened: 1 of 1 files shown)


NTaylorMullen:
Note: For some reason I can no longer run the repository so I couldn't validate end-to-end myself

dbaeumer:
I merged it and will test.

dbaeumer:
OK. I tested it and it disable it for 3.17 as well. I did revert the comment. Can you please investigate again. Thanks!

PR #1380, opened 2021-11-17
Add D language ID, fix #1379

### LINKED ISSUE #1379: Add language ID for D
The language identifier `d` has been used for various [D](https://dlang.org) (dlang) extensions on vscode already and fits right into the naming scheme. The language `D` with a mapping to the identifier `d` should be added in the list of languages in https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem (and future spec versions)

WebFreak001:
just adds D to the language ID tables that recommend file type meanings as it is a fairly established language and is already used in all D LSPs (dls, serve-d) like this and fits the other naming conventions like for `C`

Changes nothing in semantics or protocol definitions.

### DIFF
--- a/_specifications/specification-3-14.md
+++ b/_specifications/specification-3-14.md
@@ -666,6 +666,7 @@ C | `c`
 C++ | `cpp`
 C# | `csharp`
 CSS | `css`
+D | `d`
 Diff | `diff`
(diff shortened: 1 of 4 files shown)


dbaeumer:
Added by hand.

PR #1393, opened 2021-12-17
Make versions consistently have type uinteger

michaelpj:
`PublishDiagnosticParams` has its `version` field specified as a
`uinteger`, but other occurences of `version` are `integer`s.

This is inconsistent, and results in servers having to do
(theoretically) lossy conversions to `uinteger`s when creating
`PublishDiagnostics` notifications.

It seems to me that versions probably _should_ be `uintegers` (starting
at 0 and monotonically increasing), so this changes the other places to
use `uinteger`.

I suspect this will have no effect on implementations, but in theory
this could cause trouble for any clients who currently make use of
negative versions for something.

### DIFF
--- a/_specifications/specification-3-17.md
+++ b/_specifications/specification-3-17.md
@@ -1242,7 +1242,7 @@ interface TextDocumentItem {
 	 * The version number of this document (it will increase after each
 	 * change, including undo/redo).
 	 */
-	version: integer;
+	version: uinteger;
 
 	/**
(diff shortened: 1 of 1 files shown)


dbaeumer:
We have to keep this as integer to stay backwards compatible since the first version of LSP allowed negative version numbers.

dbaeumer:
I made the change to `integer`

PR #1401, opened 2022-01-19
add support for textDocument/validateBreakableRange

kayle:
This new capability allows the server to adjust a breakpoint range.

- The `textDocument/validateBreakableRange` request is sent from the client to the server to adjust a breakpoint or active 
statement range. The server response is the best corresponding range for the client to use. `null` should be returned if a 
breakpoint should not be placed near the requested range. If the server is not able to determine a precise range (e.g. due to 
the range being in a disabled preprocessor block), the server can return a 0 width range located at the beginning of the line.


_specifications/specification-3-17.md
```diff
@@ -9176,6 +9183,42 @@ export interface Moniker {
 
 Server implementations of this method should ensure that the moniker calculation matches to those used in the corresponding LSIF implementation to ensure symbols can be associated correctly across IDE sessions and LSIF indexes.
 
+#### <a href="#textDocument_validateBreakableRange" name="textDocument_validateBreakableRange" class="anchor">Validate Breakable Range (:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The `textDocument/validateBreakableRange` request is sent from the client to the server to adjust a breakpoint or active statement range. The server response is the best corresponding range for the client to use. `null` should be returned if a breakpoint should not be placed near the requested range. If the server is not able to determine a precise range (e.g. due to the range being in a disabled preprocessor block), the server can return a 0 width range located at the beginning of the line.
+
```
dbaeumer, line 9191: We need to add a client capability that indicates whether that request support dynamic registration on the client side. You can copy that from another request.

_specifications/specification-3-17.md
```diff
@@ -9176,6 +9183,42 @@ export interface Moniker {
...
+_Server Capability_:
+
+* property name (optional): `breakableRangeProvider`
+* property type: `boolean`
```
dbaeumer, line 9195: We need to add a option definition and a option registration definition. The registration definition should extend from `TextDocumentRegistrationOptions & StaticRegistrationOptions` since this is what new request are supposed to do.

dbaeumer (changes requested):

dbaeumer:
@kayle thanks for the PR. I added some comments

_specifications/specification-3-17.md
```diff
@@ -9176,6 +9183,42 @@ export interface Moniker {
 
 Server implementations of this method should ensure that the moniker calculation matches to those used in the corresponding LSIF implementation to ensure symbols can be associated correctly across IDE sessions and LSIF indexes.
 
+#### <a href="#textDocument_validateBreakableRange" name="textDocument_validateBreakableRange" class="anchor">Validate Breakable Range (:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.17.0*
+
+The `textDocument/validateBreakableRange` request is sent from the client to the server to adjust a breakpoint or active statement range. The server response is the best corresponding range for the client to use. `null` should be returned if a breakpoint should not be placed near the requested range. If the server is not able to determine a precise range (e.g. due to the range being in a disabled preprocessor block), the server can return a 0 width range located at the beginning of the line.
```
dbaeumer, line 9190: I think we should phrase this a little different since the server shouldn't adjust breakpoints. I was always under the assumption that the method is used to validate if the language server thinks that a breakpoint makes sense at the given range and if necessary expands it.

I also think that we need to write something about when clients should usually call this (e.g. a user sets a breakpoint and the debugger is not active yet; so a language server might be helpful to determine if a breakable range exist). But the language server could be wrong as well. The final instance is the debugger.

tmat: > that the method is used to validate if the language server thinks that a breakpoint makes sense at the given range and if necessary expands it.

Yes, that's correct. 

> But the language server could be wrong as well. The final instance is the debugger.

I wouldn't say the language server may be wrong. Rather, I'd explicitly distinguish between placing breakpoint spans vs binding breakpoints when attaching to process. Breakpoint placement is entirely up to the language server and its entirely dependent on syntax around the position in the document where the breakpoint is placed. Debugger has no say in that. Once the IDE is attached to the debuggee then debugger uses the breakpoint spans already placed in the source code and _binds_ them (via debugger adapter protocol). For managed languages binding means associating the breakpoint with a sequence point in the PDB. If the debugger is already attached when the breakpoint is being place then the steps are the same -- first breakpoint is placed by language service and then bound by debugger.

kayle: Thanks for the feedback. How does the updated description look? I tried to keep it somewhat general since I'm not sure all debuggers will operate exactly the same as .net does with pdbs.

dbaeumer:
@kayle the spec part looks good to me. One additional question: do you have an implementation for it. We usually require at least one implementation and it would be good if I can point to one.

dbaeumer:
@kayle cool. Will monitor how it goes and then merge this PR.

PR #1442, opened 2022-04-11
Make PositionEncodingKind a type

michaelpj:
As discussed here:
https://github.com/microsoft/language-server-protocol/commit/f9c85d535bbbf918d2433c034f20f99ddadf0ba1#r70968669

I made a change that reflected the existing state, but looking at it, it still reads somewhat ambiguously to me.
We are talking about an overall "position encoding", but in fact it only has quite a specific meaning: what is counted by character offsets.                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                      
If we were open to a slightly bigger change, we could consider something like:
```typescript
// with a proper type obviously
type CharacterOffsetUnit = 'utf-8-codeunits'|'utf-16-codeunits'|'unicode-codepoints';
interface PositionCapabilities {
  characterOffsetUnits: CharacterOffsetUnit;
}
```


_specifications/lsp/3.17/types/position.md
```diff
@@ -10,13 +10,51 @@ interface Position {
...
 	 * to the line length.
 	 */
 	character: uinteger;
 }
-```
\ No newline at end of file
+```
+
+When describing positions the protocol needs to specify how offsets (specifically character offsets) should be interpreted.
+The corresponding `PostionEncodingKind` is negotiated between the server and client and server during initialization.
```
dbaeumer, line 24: I think this should read ... is negotiated between the server and the client during initialization.

dbaeumer (changes requested):

_specifications/lsp/3.17/types/position.md
michaelpj: Yes, thanks.

michaelpj: Fixed.

michaelpj:
> Maybe I'm overthinking this, but explaining the original a𐐀b example of the current specification for each encoding would make this even clearer. Or not. Thank you.

I think so long as we're referring to clearly understood ways of dealing with text, we can assume that programmers will go elsewhere to understand what e.g. "indexing in UTF-8 code units" means.

dbaeumer:
> I think so long as we're referring to clearly understood ways of dealing with text, we can assume that programmers will go elsewhere to understand what e.g. "indexing in UTF-8 code units" means.

@michaelpj are you willing to provide this as a PR as well?

michaelpj:
I meant that I think the current text is clear enough, since it uses standard terminology. Perhaps @nemethf disagrees?

nemethf:
> Perhaps @nemethf disagrees?

I don't.

PR #1443, opened 2022-04-12
Add a note about relative links in MarkupContent

DanTup:
@dbaeumer is this what you had in mind for https://github.com/microsoft/vscode/issues/142051#issuecomment-1096205107?

I don't know if MarkupContent is currently supported anywhere that isn't tied to a request for a document, but it probably makes sense to have this caveat?


DanTup:
@dbaeumer I think in the client I want to add this in protocolConverter.ts's `asMarkdownString`, but I don't really have access to requests there. I could pass it down through the arguments, but I guess that's either a breaking change or it'd have to be optional (which would make it easy to miss from future requests).

Is there a better way/place I can implement this? (Ideally, we could do it centrally/generically, detect the request has a document URI?).

_specifications/lsp/3.17/types/markupContent.md
```diff
@@ -1,6 +1,6 @@
 #### <a href="#markupContent" name="markupContent" class="anchor"> MarkupContent </a>
 
- A `MarkupContent` literal represents a string value which content can be represented in different formats. Currently `plaintext` and `markdown` are supported formats. A `MarkupContent` is usually used in documentation properties of result literals like `CompletionItem` or `SignatureInformation`. If the format is `markdown` the content should follow the [GitHub Flavored Markdown Specification](https://github.github.com/gfm/).
+ A `MarkupContent` literal represents a string value which content can be represented in different formats. Currently `plaintext` and `markdown` are supported formats. A `MarkupContent` is usually used in documentation properties of result literals like `CompletionItem` or `SignatureInformation`. If the format is `markdown` the content should follow the [GitHub Flavored Markdown Specification](https://github.github.com/gfm/). `MarkupContent` in returned in responses to requests that are scoped to a document may contain relative links, which will be resolved relative to that document.
```
KamasamaK, line 3: I think this needs to be reworded. "in returned in responses" doesn't sound right.

DanTup: Thanks! I've removed `in returned`.. I don't know how I wrote that, or how I missed it when re-reading 🙈

DanTup:
@dbaeumer I started looking at this again today and realised that it's not as simple as I first thought. The documentation being rendered in a hover or code completion results might contain paths that are relative to the document containing the markdown, not relative to the document triggering the hover/completion (eg. you could be completing or hovering symbols from another file).

So I think this either needs to be handled in the servers (they rewrite the paths), or LSP allows the base URI to be included in MarkdownString? (the latter avoids the server having to parse Markdown).

Issue #1426, opened 2022-03-21
[folding] custom folding text for folded ranges


ris58h:
Other IDE's like `Visual Studio` or `Intellij IDEA` allow to define custom text for folded regions.

There are a lot of issues that could be solved with this feature: https://github.com/microsoft/vscode/issues/70794 , https://github.com/microsoft/vscode/issues/3352 , https://github.com/microsoft/language-server-protocol/issues/827 and other.

The proposed change is to add `collapsedText` property to [FoldingRange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#foldingRange) interface:
```
interface FoldingRange {
	startLine: uinteger
	startCharacter?: uinteger
	endLine: uinteger;
	endCharacter?: uinteger;
	kind?: string;

        collapsedText?: string;
}
```

The problem is: what should clients that don't support char-ranges folding (`lineFoldingOnly === true`) do? I've described the problem in https://github.com/microsoft/vscode/issues/70794
Possible options are:
1. Clients must replace the whole first line too.
2. Clients must show the `collapsedText` at the end of the first line.
3. Clients are free to decide how they show it.
4. Add a property to the `FoldingRange` to specify the behavior.

aeschli:
Having a collapsedText makes sense. 
Following the current logic, for lineFoldingOnly === true, by default it will be shown at the end of the line.

To replace the full line, we could allow that the startCharacter can be set to 0, even when lineFoldingOnly === true.

ris58h:
@aeschli it's a great idea to use `startCharacter` 👏
Do we need an additional property in `FoldingRangeClientCapabilities` to specify that client doesn't support `collapsedText`?

aeschli:
Yes, a new capability is needed. It will be false by default.

ris58h:
Ok. What's next? Should I make PR in https://github.com/microsoft/vscode-languageserver-node repository?

dbaeumer:
Yes, first we need to have an implementation for this.

ris58h:
The PR is ready https://github.com/microsoft/vscode-languageserver-node/pull/910

ryzngard:
@ris58h are you able to do a PR against the specification here? Definitely interested in this moving forward.

ris58h:
@ryzngard I'm interested too but I'm on vacation right now, so not in the very near future.

ryzngard:
[...]

PR #1445, opened 2022-04-20
Add collapsedText to foldingRange

ryzngard:
This proposes an additional property on foldingRange that allows a server to specify what text should be shown by the client when text is folded. 

#1426


NTaylorMullen (approved):

KamasamaK:
This is missing the client capability.

ryzngard:
> This is missing the client capability.

Added

PR #1453, opened 2022-04-27
Update DocumentDiagnosticReportKind case

KamasamaK:
Changing `DocumentDiagnosticReportKind` kinds to match the case convention used by other *-`Kind`s.


dbaeumer:
Makes perfect sense

PR #1473, opened 2022-05-17
Add lsp-bridge, lsp-bridge is LSP client for Emacs

manateelazycat:
lsp-bridge's goal is to become the fastest LSP client in the Emacs.

lsp-bridge use python threading technology build cache bridge between Emacs and LSP server, you will never feel stuck when you write the code.


manateelazycat:
Thanks @dbaeumer  ;)

PR #1479, opened 2022-05-23
Fix grammar (extra `the`)

asashour:


dbaeumer (changes requested):
Can you please exclude the metaModel.json. This file is generated from other sources. I will fix the grammar issuie there

asashour:
Done, thanks.

dbaeumer (approved):

PR #1487, opened 2022-06-01
Fix `WatchKind.type.name` to match JSON Schema

reznikmm:
If I understand JSON Schema right `uinteger` isn't allowed value of `name` property (only `string` and `integer` are allowed):
```json
    "EnumerationType": {
      "properties": {
        "name": {
          "enum": [
            "string",
            "integer"
          ],
          "type": "string"
        }
```

### DIFF
--- a/_specifications/lsp/3.17/metaModel/metaModel.json
+++ b/_specifications/lsp/3.17/metaModel/metaModel.json
@@ -13422,7 +13422,7 @@
 			"name": "WatchKind",
 			"type": {
 				"kind": "base",
-				"name": "uinteger"
+				"name": "integer"
 			},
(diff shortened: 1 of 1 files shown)


KamasamaK:
`WatchKind` is defined as `uinteger` in the spec, though. If anything, the `enum` should have `uinteger` added.

dbaeumer:
@reznikmm thanks for the PR. 

We shouldn't edit the meta model by hand. It is generated in the vscode-languageserver-node repository from the source  code. I will regenerate the model and update it.

dbaeumer:
See https://github.com/microsoft/language-server-protocol/issues/1488

PR #1504, opened 2022-06-20
SemanticTokensLegend uses enums

### LINKED ISSUE #1503: SemanticTokensLegend doesn't use enum types
```
export interface SemanticTokensLegend {
	/**
	 * The token types a server uses.
	 */
	tokenTypes: string[];

	/**
	 * The token modifiers a server uses.
	 */
	tokenModifiers: string[];
}
```
Why not
```
export interface SemanticTokensLegend {
	/**
	 * The token types a server uses.
	 */
	tokenTypes: SemanticTokenTypes[];

	/**
	 * The token modifiers a server uses.
	 */
	tokenModifiers: SemanticTokenModifiers[];
}
```
Would need a metamodel change too, I think making `SemanticTokenTypes` and `SemanticTokenModifiers` allow custom values would also be necessary (and seems accurate).

michaelpj:
Fixes #1503.

I'm happy to update the metamodel too, but I'm not sure where to do it now - in the `vscode-languageserver-node` repo or here?

### DIFF
--- a/_specifications/lsp/3.17/language/semanticTokens.md
+++ b/_specifications/lsp/3.17/language/semanticTokens.md
@@ -85,12 +85,12 @@ export interface SemanticTokensLegend {
 	/**
 	 * The token types a server uses.
 	 */
-	tokenTypes: string[];
+	tokenTypes: SemanticTokenTypes[];
 
 	/**
 	 * The token modifiers a server uses.
 	 */
-	tokenModifiers: string[];
+	tokenModifiers: SemanticTokenModifiers[];
 }
 ```
 


dbaeumer:
Actually this is on purpose due to:

> The protocol defines a set of token types and modifiers but clients are allowed to extend these and announce the values they support in the corresponding client capability

To give clients more experimentation room a client and server can agree on an additional set of token types. To support that `string[]` is used as a type.

I will close the PR. Please ping if you disagree.

michaelpj:
But how is that different from, say, `CodeActionKind`? The set of kinds is open, and negotiated between the client and the server, but we still use the `CodeActionKind` type in messages (under the understanding that it's an open enum, which is even nicely visible in the meta model). Why is it okay there but not here?

KamasamaK:
I agree that it should be able to work more like `CodeActionKind`, but that's not an actual `enum`. You'd need to redefine `SemanticTokenTypes` and `SemanticTokenModifiers` in the same way to make it work like that.

michaelpj:
In the metamodel they are both enumerations, it's just that `CodeActionKind` is declared as allowing custom values. It sounds like that means there would need to be a change on the typescript side too?

dbaeumer:
Actually we should do exactly that. A enum can't be redeclared as a string it TS.

dbaeumer:
Actually now I remember. I can't do that since `class` and `enum` can't be used as variable names. It would be breaking doing this. I will look into fixing this in the meta model generation.

michaelpj:
> It would be breaking doing this.

Not for the protocol, though? I think the protocol should be fine.

dbaeumer:
Yes, that is true. But some people parse the TS code embedded in the spec right now and that would then fail as well with the TS tooling.

michaelpj:
Well, they should use the metamodel :)

Anyway, if you know how you want to fix this I'll let you do it.

dbaeumer:
Yes, fixed it in the meta model last night. If enough people have adopted the meta model we can consider changing the spec to make things more consistent.

PR #1526, opened 2022-08-04
Clarify properties in document filter.

dbaeumer:


alexr00 (approved):

_specifications/lsp/3.17/types/documentFilter.md
```diff
@@ -37,6 +37,8 @@ export interface DocumentFilter {
 }
 ```
 
+Please note that for a document filter to be valid at least one of the properties for `language`, `scheme`, or `pattern` must be set. To keep the type definition simply all properties are marked as optional.
```
KamasamaK, line 40: This should say "simple", not "simply".

dbaeumer: Fixed.

PR #1544, opened 2022-09-08
Create deno.yml

sujadhali:


dbaeumer:
@sujadhali can you please explain why thin workflow should be added?

dbaeumer:
I will close the PR since I do not see any reason merging it.

Issue #1411, opened 2022-02-15
Documentation - links in the spec document on the website point to a different version


pedro-w:
On the LSP website, internal links for the current (3.16) version point to elements in the 3.17 version. This is minor but it's a bit tiresome when trying to note down references to a stable version. There are no such links in the markdown source so I'm not sure how they get added; apologies if this is the wrong place to report.
For example: https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#requestMessage is a location in the 3.16 spec, but just below the heading is a block of interface code, containing links to, amongst others, the 3.17 version of [RequestMessage](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#requestMessage)
Would it be possible to have the link target be the same as its source, i.e. 3.16 -> 3.16, current -> current and so on?

dbaeumer:
@NTaylorMullen can I ask you to have another look at this.

NTaylorMullen:
> @NTaylorMullen can I ask you to have another look at this.

On my todo list for a while now, promise I'll get to it eventually! 😄

pedro-w:
This still happens but it's not so bad now as the links do point to a stable version (I think they used to point to version 'current' when I reported this), so thanks for your support so far. I had a look myself, I can build the pages locally with Ruby etc. but for the life of me I can't figure out how the hyperlinks get added to the markdown code blocks. How does it work?

dbaeumer:
@NTaylorMullen friendly ping.

pedro-w:
Don't wait for the unfriendly ping :wink:

Seriously though, you guys have got more important things to worry about -  I'd be happy to have a poke around and see if I can fix it myself, but I could just do with a hint to get started !

dbaeumer:
@pedro-w thanks!!

@NTaylorMullen can you help @pedro-w to get started?

pedro-w:
OK, I think it is hard-coded in 

https://github.com/microsoft/language-server-protocol/blob/ff54ee8f15365600ee8c53e00644b231f51c9368/_includes/js_files.html#L40

and links are generated in the browser by JS in that file (that's why I couldn't find it!)
Is that correct?

pedro-w:
I think I've found a way to fix this which needs only small changes. I will put up a PR "For discussion only" and you can maybe comment if it's feasible (you may have other constraints I've not thought of). However it'll take me a couple of days as I am a bit short of time at the moment.

[...]

PR #1551, opened 2022-09-17
Fix links in the documentation

pedro-w:
This is linked to #1411. In the spec documents, certain identifiers in the code blocks are converted into hyperlinks by some Javascript which runs in the user's browser. Currently, these links are resolved to a hard-coded base URL, namely the 3.17 spec. This means that links within the 3.16 document navigate to the 3.17 spec, and also means that future revisions will either need the hard-coded link to be changed, or risk pointing back to an older version.

There are two cases:
1. In the LSP specs the URLs should be left as fragments, and will therefore link to anchors in the same document
2. In the 'associated specs' (i.e. LSIF) the URLs should be resolved by the script to the corresponding LSP spec document

I thought it would be useful to allow for more 'associated specs' to be added in the future, and that they might need to refer to a specific version of the LSP spec.

Therefore I modified the JS code to pick up the LSP version from a key named `lspVersion` in the front matter of the page source. The information to get the (relative) base URL for a particular version is already in the repository, in `_data/specifications.yml`; I needed to add a `version` key to allow the correct one to be looked up.

* In the LSP spec documents, `lspVersion` is not defined and the links are left as fragment URLs to the same document (i.e. internal links)
* In the LSIF spec documents, `lspVersion` is defined and, via `_data/specifications.yml`, the URLs are converted to relative links to the appropriate spec version.

This happens when Jekyll builds the pages and there is no additional overhead when the user views the page.

Commit bbf793c implements this.
Commit 62077f2 is optional and just uses a JS Map to store the linking information instead of a bare object. 
Commit 58f8e0f is also optional and removes some absolute URL links from the LSIF page, because the auto-linking will now target the correct files.

### DIFF
--- a/_data/specifications.yml
+++ b/_data/specifications.yml
@@ -2,8 +2,10 @@
   anchor: lsp
   children:
   - title: 3.17 (Current)
+    version: 3.17
     url: /specifications/lsp/3.17/specification
   - title: 3.16 (Previous)
+    version: 3.16
     url: /specifications/specification-3-16
 - title: LSIF
   anchor: lsif
--- a/_includes/js_files.html
+++ b/_includes/js_files.html
@@ -19,13 +19,13 @@
 
 
 <script type="text/javascript">
-  var linkableTypes = {
-  {% for linkableGroup in site.data.linkableTypes %}
-  {% for linkableType in linkableGroup.children %}
-  "{{ linkableType.type }}": "{{ linkableType.link }}",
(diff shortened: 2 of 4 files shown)


NTaylorMullen (approved):
Love your approach and really appreciate the detailed write-up on what you did / what you considered. Looks good!

dbaeumer (approved):

PR #1597, opened 2022-11-30
Add "noActiveParameterSupport" to allow null active parameters

DanTup:
See https://github.com/microsoft/vscode/issues/145851.

(@dbaeumer not sure if this exactly what you had in mind, or if the naming is reasonable)


jdneo:
@dbaeumer Do we still take this change into consideration?

dbaeumer:
Yes, I do. But I want to first ship the 8.1.0 release this week / beginning of next week and the start on a new release.

dbaeumer:
The spec change looks good to me. @DanTup will you do a PR for the implementation as well?

DanTup:
Yep! Looks like this needs some tweaks here - my comment says `SignatureHelp` but I added the comments to `SignatureInformation`. I think `null` needs supporting on both. I'll update this PR and open a PR for the implementation and make sure they match.

DanTup:
I've pushed some tweaks here, and also implementation changes to https://github.com/microsoft/vscode-languageserver-node/pull/1192.

_specifications/lsp/3.18/language/signatureHelp.md
```diff
@@ -217,15 +227,24 @@ export interface SignatureHelp {
...
+	 * `signatures[activeSignature].parameters` defaults to 0 if the active
+	 * signature has parameters.
+	 * 
+	 * If the active signature has no parameters it is ignored.
+	 * 
 	 * In future version of the protocol this property might become
-	 * mandatory to better express the active parameter if the
-	 * active signature does have any.
+	 * mandatory (but still nullable) to better express the active parameter if
+	 * the active signature does have any.
```
dbaeumer, line 245: Please specify that `null` is only supported from @since 3.18.0 on forward

_specifications/lsp/3.18/language/signatureHelp.md
```diff
@@ -258,11 +277,17 @@ export interface SignatureInformation {
...
-	 * If provided, this is used in place of `SignatureHelp.activeParameter`.
+	 * If `null`, no parameter of the signature is active (for example a named
+	 * argument that does not match any declared parameters). This is only valid
+	 * if the client specifies the client capability
+	 * `textDocument.signatureHelp.noActiveParameterSupport === true`
+	 *
+	 * If provided (or `null`), this is used in place of
+	 * `SignatureHelp.activeParameter`.
 	 *
 	 * @since 3.16.0
```
dbaeumer, line 288: Same here for `null`

dbaeumer (changes requested):

_specifications/lsp/3.18/language/signatureHelp.md
DanTup: I've added this to the comments, although I don't think I can use `@since` without it applying to the whole field?

_specifications/lsp/3.18/language/signatureHelp.md
```diff
@@ -217,15 +227,24 @@ export interface SignatureHelp {
...
+	 * signature has parameters.
+	 * 
+	 * If the active signature has no parameters it is ignored.
+	 * 
 	 * In future version of the protocol this property might become
-	 * mandatory to better express the active parameter if the
-	 * active signature does have any.
+	 * mandatory (but still nullable) to better express the active parameter if
+	 * the active signature does have any.
 	 */
```
dbaeumer, line 246: I usually do the following:

@since 3.18.0: support for `null`

Any reason why this shouldn't work

_specifications/lsp/3.18/language/signatureHelp.md
```diff
@@ -258,11 +277,17 @@ export interface SignatureInformation {
...
+	 * If `null`, no parameter of the signature is active (for example a named
+	 * argument that does not match any declared parameters). This is only valid
+	 * since 3.18.0 and if the client specifies the client capability
+	 * `textDocument.signatureHelp.noActiveParameterSupport === true`
+	 *
+	 * If provided (or `null`), this is used in place of
+	 * `SignatureHelp.activeParameter`.
 	 *
 	 * @since 3.16.0
 	 */
```
dbaeumer, line 289: Same as above.

dbaeumer (changes requested):

_specifications/lsp/3.18/language/signatureHelp.md
DanTup: Oh, I thought `@since` was some special tag being parsed that could only appear once on a field. I'll add this then :-)

DanTup: @dbaeumer is it intended that this would happen in the case where there isn't another `@since`?

Screenshot 2023-03-06 at 16 35 34

When there's another, then it looks like this:

Screenshot 2023-03-06 at 16 37 02

There's some similar instances in the existing meta model:

https://github.com/microsoft/vscode-languageserver-node/blob/4d7d8f4229a539f36bc28c634470539bcda8b362/protocol/metaModel.json#L4680

I don't know if `"since"` is expected to be parseable as a version number?

_specifications/lsp/3.18/language/signatureHelp.md
dbaeumer: I think what we need to do is to support more than one since tag in the meta model.

dbaeumer: Created https://github.com/microsoft/vscode-languageserver-node/issues/1204.

dbaeumer (approved):

connor4312 (approved):

PR #1602, opened 2022-12-06
Support for evaluableExpression

### LINKED ISSUE #1596: Some `evaluatableExpression: Range -> { string, Range }` to know what expression is evaluable at a given location
The story is: assuming I have a piece of JS code `myObject.myProperty`, I want that when hovering on `myProperty`, the debugger gets a change to evaluate `myObject.myProperty`.
And yes, on 2nd thought, it seems related to LSP more than DAP.

This is a followup of https://github.com/microsoft/debug-adapter-protocol/issues/343#issuecomment-1327899425 , it looks like the Debug Adapter Protocol actually doesn't know anything about the language, just like the client. So in general the Debug Adapter won't be capable of expanding a position or a selection to the right expression to resolve.
Inferring the evaluatable expression requires knowledge of the language, requires the AST. So LSP should be the brick capable of providing it.

mickaelistria:
Fixes https://github.com/microsoft/language-server-protocol/issues/1596

### DIFF
--- a/_specifications/lsp/3.17/general/initialize.md
+++ b/_specifications/lsp/3.17/general/initialize.md
@@ -871,6 +871,12 @@ interface ServerCapabilities {
 	workspaceSymbolProvider?: boolean | WorkspaceSymbolOptions;
 
 	/**
+	 * The server has support for evaluable expression.
+	 *
+	 * @since 3.18.0
+	 */
+
+	/**
 	 * Workspace specific server capabilities
 	 */
 	workspace?: {
--- a/_specifications/lsp/3.18/language/evaluableExpression.md
+++ b/_specifications/lsp/3.18/language/evaluableExpression.md
@@ -1,0 +1,76 @@
+#### <a href="#textDocument_evaluableExpression" name="textDocument_evaluableExpression" class="anchor">Evaluable Expression Request (:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.18.0*
+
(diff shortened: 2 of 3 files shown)


dbaeumer:
@mickaelistria thanks for the PR. Are you working on some sort of implementation as well. We usually require at least on implementation for an editor before we add something to the specification?

mickaelistria:
I didn't start an implementation for this proposal but will try to do it. However, please be patient as it's not trivial and I get a bunch of higher priority tasks to deal with first (and probably some more later); but if anyone likes to proposal and wants to start drafting an implementation, that would be more than welcome!

dbaeumer:
I will be patient :-)

_specifications/lsp/3.17/general/initialize.md
```diff
@@ -870,6 +870,12 @@ interface ServerCapabilities {
 	 */
 	workspaceSymbolProvider?: boolean | WorkspaceSymbolOptions;
 
+	/**
+	 * The server has support for evaluable expression.
+	 *
+	 * @since 3.18.0
+	 */
+
```
puremourning, line 878: should there be a declaration here?

KamasamaK: Not here. This is under the 3.17 spec.

dbaeumer:
@mickaelistria I will close the PR for now. If you want to continue working on it please ping (e.g. the implementation)

PR #1668, opened 2023-02-13
Fixes #1666: The term "falsy" in ColorPresentation and CompletionItem is JS/TS specific.

### LINKED ISSUE #1666: The term "falsy" in ColorPresentation and CompletionItem is JS/TS specific.
The term `falsy` is used in `ColorPresentation` once and `CompletionItem` three times, but this term is JavaScript/TypeScript specific. Other places in the spec don't use this terminology.

It would be better in these cases to use "missing", "absent", "not set", or similar instead, as these properties are optional and not nullable.

dbaeumer:
Fixes #1666


KamasamaK:
Is "omitted" the equivalent to "falsy" in all of these cases? I would expect for the `string`s it would be "omitted or empty".

aeschli (approved):

KamasamaK:
Okay, I guess not. I'll modify our SDK then.

Issue #1529, opened 2022-08-04
Support Inline Completions


KamasamaK:
The [Inline Completions API](https://github.com/microsoft/vscode/issues/124024) has been finalized in VS Code, so support should be added to the LSP.

c-claeys:
This is in the backlog, are you open to PRs for this? The public VSCode API provides a clear path, in my opinion. Happy to send PRs here for the spec and https://github.com/microsoft/vscode-languageserver-node

mroch:
Meta is also interested in helping push this forward! How can we help?

bolinfest:
@dbaeumer In implementing this in our LSP, we are finding the need to throttle these events, wait to see if a cancellation is sent, etc. As such, it would be helpful to formalize the LSP request so it takes a `VersionedTextDocumentIdentifier` (or perhaps to be more flexible, an `OptionalVersionedTextDocumentIdentifier`) rather than a plain `TextDocumentIdentifier` so that by the time the inline completions request is handled, it can verify that the version of the document it has in memory matches the one that was specified in the inline completions request (as text document update events may have been processed inbetween).

dbaeumer:
PRs are welcome. But they have to be made for `vscode-languageserver-node` (or another implementation) as well.

dbaeumer:
@bolinfest IMO this is not necessary since the ordering in which requests and notifications are sent to the server guarantees that a request is always on the latest updated document state the client has sent. If a server reorders these (which he is allowed to do) then the server needs to keep track of the versions the reordering has happened.

bolinfest:
@dbaeumer Another request (which is perhaps strictly spec-related) is that I see that [`ErrorCodes.ServerCancelled`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseMessage) has the following caveat:

```
/**
 * The server cancelled the request. This error code should
 * only be used for requests that explicitly support being
 * server cancellable.
 *
 * @since 3.17.0
 */
 ```
 
 Could we formally declare that inline completions are server cancellable?

dbaeumer:
No, we can't since this is client specific. There is a `serverCancelSupport` capability for those requests that support this. To my knowledge VS Code doesn't have cancel support (that retriggers the request) for inline completion. It does for semantic tokens.

dbaeumer:
This got added as proposed for 3.18.

PR #1673, opened 2023-02-15
Add textDocument/inlineCompletion to 3.18

c-claeys:
This is an initial pass per the discussion in https://github.com/microsoft/language-server-protocol/issues/1529 I primarily modeled it after the existing VSCode API. I'm opening a PR for `vscode-languageserver-node` in parallel.


_specifications/lsp/3.18/metaModel/metaModel.json
```diff
@@ -1,6 +1,6 @@
 {
```
dbaeumer, line 1: This we don't need to update. I have a tool that does this off the vscode-languageserver-node repository.

dbaeumer (changes requested):

_specifications/lsp/3.18/metaModel/metaModel.json
c-claeys: I ran the npm script from the other PR and cp'd it into this one (needed to change the top-level 3.17->3.18). Let me know if you'd rather I revert it.

c-claeys: I ran the npm script from the other PR and cp'd it into this one (needed to change the top-level 3.17->3.18). Let me know if you'd rather I revert it.

_specifications/lsp/3.18/language/inlineCompletion.md
```diff
@@ -0,0 +1,233 @@
+#### <a href="#textDocument_inlineCompletion" name="textDocument_inlineCompletion" class="anchor">Inline Completion Request (:leftwards_arrow_with_hook:)</a>
+
+> *Since version 3.18.0*
+
+The inline completion request is sent from the client to the server to compute inline completions for a given text document either explicitly by a user gesture or implicitly when typing.
+
+_Client Capability_:
```
rchl: This seems to tell very little about what this feature exactly is about.

Is it an alternative to normal completions or does it enhance it?
How does it differ from normal completions?
What problem does it solve that normal completions don't?

c-claeys: Added some high level description in the .md and tried to answer your questions. I hope that helps, please let me know what you think.

c-claeys:
Sorry for any spam on the review requesting here. I think it's ready for @dbaeumer and @rchl to both look at.

_specifications/lsp/3.18/language/inlineCompletion.md
rchl: That are some questions that pop into my head when trying to think about it from the implementation perspective but since I'm not implementing it, I will not create more noise about it here.

dbaeumer:
Needs to wait until next week. We have endgame week.

_specifications/lsp/3.18/language/inlineCompletion.md
```diff
@@ -2,7 +2,25 @@
...
+feature to work *with* completions presented in the IntelliSense interface.
+
+When multiple inline completion items are returned, the client may decide
+whether the user can cycle through them or if they, along with their
+`filterText`, are merely for filtering if the user continues to type without yet
+accepting the inline completion item.
+
+Clients may choose to send information about the user's current completion
+selection via context if completions are visible at the same time. In this case,
+returned inline completions should extend the text of the provided completion.
```
rwols: In what way is this different from a client requesting `textDocument/completion` and rendering the results as "inline" anyway?

What does it matter to a server whether the completions are rendered inline or with a traditional widget box?

bolinfest: I could be wrong, but one thing that seems unique to *inline* completions is that they can coexist with traditional "autocomplete suggestions," which are effectively passed along via `selectedCompletionInfo` in the VS Code API today. While the particulars of the UI implementation should be left to the client, somehow it's important to capture that these are two distinct things and that one can include the context from the other when deciding how to respond to the request?

c-claeys:
Hey Dirk, with https://github.com/microsoft/vscode-languageserver-node/pull/1190 approved, is there anything in particular you'd like in this PR?

mfussenegger (commented):
Couldn't this just be some kind of presentation hint on the completion items of the existing completion endpoint?

From a client implementation perspective, I struggle seeing advantages of having to make two requests to get two different kind of completions - especially since the spec says that the two kinds should kinda complement each other.

I can also imagine that for servers it's not really an advantage either? Given that it's possible that you have to do similar kind of computation to produce either, so you may end up doing the same computation twice, unless you add synchronization mechanisms.

dbaeumer:
Conceptually there is a difference between them. Code completions should usually only suggest things that are correct (e.g. complete an existing method, ...) were as inline completions can be way more fuzzy. That is why they complement each other. 

On the server side I think servers should already handle cases where they are ask questions for the same document in separate requests and do some AST / binding caching. If we encounter that this is a problem I would rather allow for request batch then not having two requests.

mfussenegger:
> Conceptually there is a difference between them. Code completions should usually only suggest things that are correct (e.g. complete an existing method, ...) were as inline completions can be way more fuzzy. That is why they complement each other.

How about adding a fuzzy score or confidence score to the completion results then? The `inline` in the new proposal implies how clients should present it, and so far the policy was that presentation is mostly left up to the clients. It also doesn't communicate this semantic difference you mentioned.
The current description reads a lot like regular completion if it weren't for the presentation details - and even those have overlap to how a client could already present the existing completion results 


> With only one item being displayed at a time, the server may use this
feature for visible edits where it is **relatively confident the user would like
this edit in particular**.

Isn't this contradicting that inline completions can be way more fuzzy?

dbaeumer:
I see the point that the spec should talk about how a client presents inline completions (e.g. for example as grayed text in the editor itself). I will think about how to rephrase this.

However there is a conceptual different that we would like to keep separate mostly because adding additional flags or context information always turned out to be more complicated.

>> With only one item being displayed at a time, the server may use this
feature for visible edits where it is relatively confident the user would like
this edit in particular.

> Isn't this contradicting that inline completions can be way more fuzzy?

No, completion items should for example not complete incorrect code (e.g. propose a method that doesn't exist). Inline completions are allowed to generate code that is both syntactically and semantically incorrect.

_specifications/lsp/3.18/language/inlineCompletion.md
dbaeumer: I rephrased this since @mfussenegger is correct that we shouldn't mention any required UI implementations in the specification.

```
The inline completion request is sent from the client to the server to compute inline completions for a given text document either explicitly by a user gesture or implicitly when typing. 

Inline completion items usually complete bigger portions of text (e.g. whole methods, ...) and in contrast to completions items can complete code that might be syntactically or semantically incorrect. 

Due to this inline completion items are usually not suited to be presented in normal code completion widgets like a list of items. One possible approach can be to present the information inline in the editor with lower contrast.

When multiple inline completion items are returned, the client may decide whether the user can cycle through them or if they, along with their `filterText`, are merely for filtering if the user continues to type without yet accepting the inline completion item.

Clients may choose to send information about the user's current completion selection via context if completions are visible at the same time. In this case, returned inline completions should extend the text of the provided completion.
```

dbaeumer (changes requested):
See my comment about the UI implementation mentioning.

_specifications/lsp/3.18/language/inlineCompletion.md
c-claeys: Phrasing looks good to me.

c-claeys:
(sorry about the churn)

dbaeumer (approved):

bpasero (approved):

PR #1679, opened 2023-02-21
Clarify requirement of `file` scheme to be a file written on disk

### LINKED ISSUE #1676: Clarify requirement of `file` URI scheme to be written to disk
This is a follow up from the conversion in https://github.com/microsoft/vscode-eslint/pull/1560 regarding unwritten buffers and the requirement of the `file` URI scheme to be written to disk.  VS Code uses the `untitled` scheme, and from the conversation in the mentioned PR, the clarification should be focused not on how clients should specify untitled/unwritten buffers, but rather that the `file` scheme should **only** be used if a file is written to disk.

I'm happy to open PR to update the spec, but I'd like to know your preference of where this would fit. I'm thinking perhaps a short mention of this in the [URI section](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri)? Or is there a better place to document this?

mskelton:
Fixes #1676

This PR clarifies that a URI with a scheme of `file` must be a file that is written on disk. Documents not written to disk (untitled or scratch files) are not given explicit instructions as to what scheme they should user, other than the fact that they should use a non-`file` scheme.

### DIFF
--- a/_specifications/lsp/3.17/types/uri.md
+++ b/_specifications/lsp/3.17/types/uri.md
@@ -27,3 +27,5 @@ There is also a tagging interface for normal non document URIs. It maps to a `st
 ```typescript
 type URI = string;
 ```
+
+Please note that a URI with a scheme of `file` represents a file that is written on disk. Documents that exist in the editor but are not written on disk, sometimes called untitled or scratch files, must used a different scheme which can be decided by the client. If a document is deleted on disk but still present in the editor, the URI may retain the scheme of `file` as some editors allow the user to save the document on disk using the same name.
--- a/_specifications/lsp/3.18/types/uri.md
+++ b/_specifications/lsp/3.18/types/uri.md
@@ -27,3 +27,5 @@ There is also a tagging interface for normal non document URIs. It maps to a `st
 ```typescript
 type URI = string;
 ```
+
+Please note that a URI with a scheme of `file` represents a file that is written on disk. Documents that exist in the editor but are not written on disk, sometimes called untitled or scratch files, must used a different scheme which can be decided by the client. If a document is deleted on disk but still present in the editor, the URI may retain the scheme of `file` as some editors allow the user to save the document on disk using the same name.


_specifications/lsp/3.17/types/uri.md
```diff
@@ -26,4 +26,6 @@ There is also a tagging interface for normal non document URIs. It maps to a `st
 
 ```typescript
 type URI = string;
-```
\ No newline at end of file
+```
+
+Please note that a URI with a scheme of `file` represents a file that is written on disk. Documents that exist in the editor but are not written on disk, sometimes called untitled or scratch files, must used a different scheme which can be decided by the client. If a document is deleted on disk but still present in the editor, the URI may retain the scheme of `file` as some editors allow the user to save the document on disk using the same name.
```
puremourning, line 31: The problem with this statement is that it is contradictory.

If the file is not on disk, don't use `file:`
Unless the file may-have-previously-or-may-in-the-future exist on disk, in which case, do.

How are client authors supposed to implement that?

Also, not providing guidance on what client authors _should_ do in the case of unsaved-but-named, or unsaved-but-unnamed files will lead to massive divergence. Having this "clarification" only seems to make things _less_ clear.

rchl: Would it be fair to say that if the client has already taken the ownership of the file by triggering `textDocument/didOpen` then it's fine to continue using the `file:` scheme, even if file is removed, but if it hasn't opened the file yet and it's about to then at that point it should not use the `file` scheme?

At least that's how I understand @dbaeumer 's explanation.

dbaeumer: IMO it is more about the intend than the current state. If a client creates a scratch buffer which intend is usually to not store the file on disk the client should not assign a file scheme but use a different scheme which IMO can be left to the client. IMO using the file scheme creates confusion especially if some other tool will put a file on disk with the same path.

If a client creates a file and doesn't immediatelly store it on disk but its intend is to do so or if the file existed on disk and got deleted it is IMO OK to use a file scheme or to keep the file scheme.

mfussenegger: > If a client creates a file and doesn't immediatelly store it on disk but its intend is to do so or if the file existed on disk and got deleted it is IMO OK to use a file scheme or to keep the file scheme.

This is the default operation mode in neovim and several language servers have problems with that. That's what led to this PR.

I fully agree with @puremourning that the definition in this PR is contradicting.

I also still think that saying "use any scheme other than `file`" for files not yet stored on disk is going to cause problems. I wouldn't be surprised if some language servers already have some `scheme == "untitled"` handling today.

dbaeumer: The problem is that there can be many schemes for things that don't exist on disk. For example VS Code has a special them for files that come from git or from Github. Since we can't enumerate and define them all IMO we can only define that URI with scheme file are intended to be on disk. Server should still handle the case that they might disappear but they IMO could make assumptions about the path and for example into which project such a file should belong to. Scratch buffers usually are different in that (e.g. they usually belong to a virtual project). This is why I think they shouldn't be using a file URI.

Although even I did in ESLint :-( I agree servers shouldn't do checks like `scheme == "untitled"`. I rewrote my checks into `scheme !== "file"` which ensure that I treat all files that are not on disk like they are virtually belonging to the workspace root (if one exists)

mfussenegger: > The problem is that there can be many schemes for things that don't exist on disk. For example VS Code has a special them for files that come from git or from Github. Since we can't enumerate and define them all IMO we can only define that URI with scheme file are intended to be on disk

I see, that's a fair point.


> Scratch buffers usually are different in that (e.g. they usually belong to a virtual project). This is why I think they shouldn't be using a file URI.

But in the Neovim case we're not talking about scratch buffers, it's the default flow to create a file

- User creates a new buffer with a file name by using `:e my_new_file_name`
- The language server client (if configured) sends a `didOpen`, using the file scheme and expected file name that the file _is_ going to have if it's saved
- User can save the file with `:w`

Given that servers should handle the case that a file might disappear. Shouldn't they also handle this use-case? An option for language servers could be to retry operations that require file access on a `didSave` if the file was missing on `didOpen`.

rwols: I don't think this is specific to Neovim. In ST we also have the flow of the user creating a new scratch buffer. We use the `buffer://` scheme for that with a `textDocument/didOpen`. Once the user hits CTRL+S to save the temporary buffer and assigns it a filename, we notify `textDocument/didClose` for the `buffer:` document, followed by a `textDocument/didOpen` for the new `file://` document on-disk.

mfussenegger: We recently made this change in neovim but it turns out this is causing issues with other language servers.

For example clangd will report errors on completion or signature help:

```
clangd: -32602: failed to decode textDocument/signatureHelp request: clangd only supports 'file' URI scheme for workspace files at (root).textDocument.uri
```

The same thing works fine in clangd with a `file` scheme even if the file is not yet saved to disk.

I'm inclined to go ahead and revert the change as it seems way more fragile (given that the file could disappear) and introduces additional complexity.

How about instead adding a blurb to `didOpen` that says that it's not guarenteed that the file exists on disk, and if it doesn't a server should retry on `didSave`.

Disk errors could happen, git checkout could remove the file. Having a client account for that and keep it consistent with each request that uses a `URI` is error prone and adds lots of overhead as you need to check each time if the file still exists. There's always a race between the clients view and the servers attempt to read the file.  For an ideal experience a server needs to recover from file read errors in any case

mskelton: @mfussenegger Thanks for the info. I'm personally fine with that approach which is what I had initially proposed in the vscode-eslint project but then I received pushback from the author with the suggestion of this direction. I do agree that using a different scheme is brittle and could easily cause race conditions and having servers handle if a file does not exist on disk would be a much more robust way of handling all possible scenarios. It doesn't preclude having early checks against the scheme as used in vscode to let the server know it doesn't even need to try checking if the file is on disk.

@dbaeumer Thoughts?

dbaeumer: I can only re-iterate what I have said before:

- if the intention is to put a file on disk then it is OK to use the `file` scheme and the file should have a reasonable path so that servers can make decisions for example to which project the file belongs.
- a file with a file scheme can be deleted from disk and it IMO can keep its URI
- if the intent is to have a scratch buffer using a file scheme is IMO a bad idea.

The ESLint server for example treats non file URIs different in terms of resolving the eslint config file used. This is why IMO a client should make a careful decision about the scheme it uses.

In general URIs can have all sort of custom themes and a client should only synchronize those documents to a server that the server can handle. To leave this up to the server to decide LSP added the concept of dynamic registration.

mskelton: > a file with a file scheme can be deleted from disk and it IMO can keep its URI

I think this is part of the issue. What I noticed with the ESLint server which led to my original PR was that when I would open a scratch buffer, the server would crash and I would have to restart the server to make linting work again. In this scenario, even if Neovim used a non-file scheme for scratch buffers, the server would crash if the file was deleted on disk but retained the `file` scheme. This is why I think having the ESLint server handle the exception more gracefully would be arguably a better fix.

dbaeumer: I think the overall confusion start with the term `scratch` buffer. I think neovim is not creating a scratch buffer. If I understand the flow correctly is that in neovim the user even already gives the file a name when a new file is created. Is this correct? This is different than in VS Code where a new file has no location and no name, hence we treat them as scratch buffers (untitled files).

I think we should still make this clear in the specification. And I also agree that a server (including ESLint) should handle file URIs that don't exist on disk yet/anymore gracefully.

mskelton: > If I understand the flow correctly is that in neovim the user even already gives the file a name when a new file is created. Is this correct?

Yeah, so here is an example flow:

A user creates a new "buffer" using a path on the file system (`:edit ~/dir/my-file.ts`). If this is a file on disk, neovim will load the contents into the buffer. If it doesn't exist on disk, it will be an empty file that is not written yet, but is aware of the file path. The user can now save the file (`:write`) which will write it to disk.

> This is different than in VS Code where a new file has no location and no name, hence we treat them as scratch buffers (untitled files).

Yeah, in Neovim you can have both. Buffers without a file path (created using `:enew) which is called an unmamed buffer but you can also as I explained in the previous paragraph create named buffers that may or may not point to a file on disk.

> And I also agree that a server (including ESLint) should handle file URIs that don't exist on disk yet/anymore gracefully.

I'll re-open my PR in vscode-eslint.

> I think we should still make this clear in the specification

@mfussenegger Any thoughts on this thread with the additional info that was posted. Is there anyway to still accomplish the goal of this spec change without causing issues to clients like Neovim?

mfussenegger: > Any thoughts on this thread with the additional info that was posted. Is there anyway to still accomplish the goal of this spec change without causing issues to clients like Neovim?

I think something like "Scratch buffers which are never meant to be saved on disk shouldn't use the `file` scheme. The `file:` scheme is reserved for URIs which are either already persisted on disk or which the client intends to save to disk. Servers must handle the case that files are intermediately not accessible and should retry on the next `didOpen` or `didSave` command" would suit neovim fine.

kaddkaka:
@mskelton was there a decision and is the spec updated now? This seems to have been closed without a comment?

mskelton:
@kaddkaka The fix for vscode-eslint was here: https://github.com/microsoft/vscode-eslint/pull/1641

The spec didn't change given that the change I suggested was not implemented. `file://` does not guarantee the file is on disk.

PR #1687, opened 2023-03-06
Fixes #1686: Clarification needed around type of params for telemetry/event

### LINKED ISSUE #1686: Clarification needed around type of params for telemetry/event
On one hand, the Language Server Protocol specification defines [notification messages](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notificationMessage) (in its subset of the JSON-RPC 2.0 spec) like so:

```ts
interface NotificationMessage extends Message {
	/**
	 * The method to be invoked.
	 */
	method: string;

	/**
	 * The notification's params.
	 */
	params?: array | object;
}
```

But further down, the definition for the [`telemetry/event`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event) notification claims:

> _Notification:_
>
> * method: ‘telemetry/event’
> * **params: ‘object’ | ‘number’ | ‘boolean’ | ‘string’;**

This seems to be a contradiction. How can `params` be `object | number | boolean | string` when the specification clearly states that notification messages may only contain arrays and objects as `params`? I'd love it if someone could clarify. Thanks so much!

dbaeumer:
Fixes #1686

### DIFF
--- a/_specifications/lsp/3.17/telemetry/event.md
+++ b/_specifications/lsp/3.17/telemetry/event.md
@@ -4,4 +4,4 @@ The telemetry notification is sent from the server to the client to ask the clie
 
 _Notification_:
 * method: 'telemetry/event'
-* params: 'object' \| 'number' \| 'boolean' \| 'string';
+* params: 'object' \| 'array';
--- a/_specifications/lsp/3.18/telemetry/event.md
+++ b/_specifications/lsp/3.18/telemetry/event.md
@@ -4,4 +4,4 @@ The telemetry notification is sent from the server to the client to ask the clie
 
 _Notification_:
 * method: 'telemetry/event'
-* params: 'object' \| 'number' \| 'boolean' \| 'string';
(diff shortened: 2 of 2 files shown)


ebkalderon:
Awesome stuff! Very happy to see this change land. To your knowledge, are there any extant language servers in the wild that use non-`object | array` values for `telemetry/event`? If this is rare, then hopefully the change from `any` to `object | number | boolean | string` in 3.16, and then from that to `object | array` in 3.17 and 3.18 won't affect too many users.

aeschli (approved):

dbaeumer:
I think that this is actually very rare.

ebkalderon:
@dbaeumer That's a relief! In that case, I'll open a few pull requests against [`lsp-types`](https://github.com/gluon-lang/lsp-types) and a few other libraries to reflect the tightening of the spec.

PR #1694, opened 2023-03-22
fix sidebar hrefs being wrong

czipperz:
I believe this is a copy/paste error.


dbaeumer:
The fix actually breaks the side bar expand /collapse behavior. If we remove it we need to adapt this as well: https://insiders.vscode.dev/github/microsoft/language-server-protocol/blob/86eba300d9c5401cfb0325562cf489168b4700f7/_layouts/specifications.html#L87

You can test the web site locally using ` bundle exec jekyll serve`

dbaeumer (changes requested):
See latest comment

dbaeumer:
Closing the PR due to inactivity.

PR #1697, opened 2023-03-22
Added Pascal and Delphi language identifier

david-millington:
This change adds both Delphi and generic Pascal languages to the list with `pascal` as the identifier for both.

Delphi has supported LSP since early 2020, and the Delphi IDE and language server use `pascal` as its identifier.

Other variants of Pascal, such as FreePascal, GNU Pascal etc likely would use the same. There is also an existing VSCode LSP plugin called OmniPascal which provides LSP for both Delphi and general Pascal and it too uses the same `pascal` identifier.

### DIFF
--- a/_specifications/lsp/3.18/types/textDocumentItem.md
+++ b/_specifications/lsp/3.18/types/textDocumentItem.md
@@ -42,6 +42,7 @@ C# | `csharp`
 CSS | `css`
 Diff | `diff`
 Dart | `dart`
+Delphi | `pascal`
 Dockerfile | `dockerfile`
 Elixir | `elixir`
 Erlang | `erlang`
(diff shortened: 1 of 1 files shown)


david-millington:
Hello. Is there anything I can do to assist this merge? I have merged into this project rarely (only once before?) and am not sure from the status (merging blocked, pending review) if this one is waiting on anything from me or anything I can help with. In terms of review, if there are any questions please let me know.

Cheers!

Wosi (approved):

dbaeumer:
Actually I forgot about this. I need to co-ordinate that a little with VS Code as well. Will look into it.

dbaeumer:
Sorry for the delay. @Wosi I looked at omni pascal and it uses `objectpascal` as the id. Are you OK with `pascal` as an id?

dbaeumer:
Seeing that @wosi approved this so going ahead with pascal.

dbaeumer:
Added by hand.

PR #1698, opened 2023-03-22
Define separate base protocol

MariaSolOs:
Questions:
1. Do we want to keep the `LSP` prefix in the base types? E.g.: [`LSPAny`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lspAny).
1. What should we do with the `since @3.X` annotations?
1. What other definitions regarding capabilities do we want to move over?


dbaeumer:
Great start. Here are my first thoughts:

- may be we should simply call it base protocol and not tools. I start to like that name better. 
- we should start with base protocol version 0.9 to make clear this is under development
- we should use BaseAny, .....
- we should remove the since annotations
- for client and server capabilities we should make examples in the base protocol, but we can't define any kind of capabilities. What we should add is a table with property names that new protocols will use in the capabilities and that LSP will not use these names top level anymore (e.g. for example a top level property `testing` for a testing protocol)

We would also need some good introduction why we are doing this. But that can what until we have all the parts in place.

For LSP 3.18 we should leave things how they are (naming wise) but ensure that we are structurally equal to the base definitions. In the LSP spec we say that for 3.19 or 3.20 we will move over to the base types but since the base concepts are structurally equals they can already be merged into one server.

What do you think?

dbaeumer:
BTW you know that you can compile the site locally as well using `bundle exec jekyll serve`

rchl:
I want to comment on `BaseAny`/`LSPAny`. Will the LSP spec then be changed to use `BaseAny` everywhere where it's using `LSPAny`? Because otherwise I think that when using types that were auto-generated from the spec, the type checkers might consider those incompatible which I'm not sure if would be intended.

EDIT: Maybe I've commented too fast. I guess your `For LSP 3.18 we should...` paragraph answers that.

dbaeumer:
The idea is to have something like `type LSPAny = BaseAny`. I would not remove `LSPAny` totally.

MariaSolOs:
> BTW you know that you can compile the site locally as well using `bundle exec jekyll serve`

@dbaeumer yeah, and I'm using the `--livereload --incremental` flags for seeing the changes live without recompiling or reloading the page ;)

But also, that's the reason why I added [this](https://github.com/microsoft/language-server-protocol/blob/c27ec2dce23057f66969f23c35add4f58ecd3c07/resources/notes.md?plain=1#L8). To prepare my development environment I had to install Ruby, install the dependencies in the Gemfile, and I also had to add `gem "webrick"` to the Gemfile to successfully build the project. I couldn't find anything in this repo that went over these steps.

MariaSolOs:
> For LSP 3.18 we should leave things how they are (naming wise) but ensure that we are structurally equal to the base definitions.

@dbaeumer I'm not sure I understand what you mean with this. Should I then not remove the base types from the 3.18 specs and rewrite them using their existing names?

dbaeumer:
I would leave the LSP spec for 3.18 like it is and we do the base protocol as a copy from approach. Of course for the base protocol we rename the type names from `LSP*` to `BaseProtocol*`

Does that make sense?

dbaeumer:
Great work. I looked over hit and here are some things we need to discuss to decide on how to move things forward:

- in the base protocol we should when possible not use examples from LSP. I think that will be confusing for people. If we can't come up with a non LSP case then we should make it explicit why we use it.
- we should remove all deprecated stuff from the base protocol (e..g rootPath, ...)
- since we actually send WorkspaceFolders in the intialize request we do need to add http://localhost:4000/language-server-protocol/specifications/lsp/3.18/specification/#workspace_workspaceFolders and http://localhost:4000/language-server-protocol/specifications/lsp/3.18/specification/#workspace_didChangeWorkspaceFolders to the base protocol.
- we should remove all @since tags as well.
- we need to add more around client & server capabilities in the base protocol. To avoid name clashes the base protocol needs to have a list of `official` take property names.

zezo2019 (approved):

MariaSolOs:
@dbaeumer I've replaced the LSP-specific examples by more general ones, specified the list of reserved capabilities and error codes, and merged some of the latest changes from the main branch. Let me know what you think 😄

dbaeumer (commented):
The enumeration section sill has a reference to a CompletionItemKind example. My be we should make a more generic enumeration here as well.

dbaeumer (commented):
The documentation of the initialize request has a `ClientCapabilities` type in the `InitializeParams`. I think we should define an empty `ClientCapabilities` best were we discuss that certain property names in the capabilities are taken by LSP.

dbaeumer (commented):
The initialize request returns a result, which basically contains the server capabilities. We should document this in the base protocol as well. See the response section here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

dbaeumer (commented):
In the register capability section I would remove the reference to TextDocumentClientCapabilities

dbaeumer (commented):
In general there are a lot of references to text documents in the register and unregister section. Since we don't have text documents in the base protocol we should think about a different example. We can also leave it quite abstract.

dbaeumer (commented):
We should also write a motivation section at the beginning explaining why we decide to specify a base protocol. I think it will make it a lot easier to understand what we want to achieve.

dbaeumer (changes requested):
See comments above.

MariaSolOs:
> The documentation of the initialize request has a `ClientCapabilities` type in the `InitializeParams`. I think we should define an empty `ClientCapabilities` best were we discuss that certain property names in the capabilities are taken by LSP.

Will do. I'll also remove the `workspaceFolders` property since we're not specifying workspace details here for now.

MariaSolOs:
> The initialize request returns a result, which basically contains the server capabilities. We should document this in the base protocol as well. See the response section here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

Oh, I had not noticed that I was missing the response. Good catch!

MariaSolOs:
> In general there are a lot of references to text documents in the register and unregister section. Since we don't have text documents in the base protocol we should think about a different example. We can also leave it quite abstract.

I think we should leave it open to the implementors to specify. After all, one of the nice things about dynamic registration is that they don't need to follow the structure from capabilities included in `initialize`, so I think leaving them abstract is good.

_specifications/base/0.9/specification.md
```diff
@@ -0,0 +1,904 @@
...
+## <a href="#baseProtocol" name="baseProtocol" class="anchor"> Base Protocol </a>
+
+The purpose of the base protocol is to create an abstraction of common editor extensibility patterns into its own specification, independently of the presence of a language server powering such extension. In particular, concepts such as server and client capabilities, the initialization and shutdown request, and the structure of requests and notifications, were originally part of the language server protocol, but there is nothing about them that should be exclusive to language service extensions.
+
+A motivating example is the [Build Server Protocol](https://build-server-protocol.github.io/). As its specification describes, all notifications and requests are defined using [the base definitions](#basicJsonStructures) of LSP, and messages such `InitializeBuild`, `OnBuildInitialized`, and `OnBuildExit` all match almost exactly with their LSP counterparts. By implementing the base protocol, a server can support multiple protocol specifications without having to reimplement the "common boilerplate" that all of them share.
+
+### <a href="#headerPart" name="headerPart" class="anchor"> Header Part </a>
```
dbaeumer, line 20: I like the motivation section. One clarification we should add is the following: the base protocol is experimental / preview and depending on the feedback we receive it might not be removed again.

I want to make sure that we don't have to maintain it in cases it is not adopted or used by other specific protocols.

dbaeumer (changes requested):
The comment above

dbaeumer (approved):

aeschli (approved):

PR #1772, opened 2023-07-17
Make includes version aware

dbaeumer:


dbaeumer:
@MariaSolOs FYI

alexdima (approved):

MariaSolOs:
@dbaeumer why was this needed? My original intent was for the `includes` files to share common definitions, while leaving the version-specific content next to the specification file (in [this folder](https://github.com/microsoft/language-server-protocol/tree/3795ff50f51545136004129489b92aa0c55625ec/_specifications/lsp/3.18)).

dbaeumer:
@MariaSolOs  Since I wanted to change something for the upcoming 3.18 spec which shouldn't be reflected in the 3.17 spec. I could have copied it down to the 3.18 spec, however that makes things somehow inconsistent (some files come from include some don't).

The thing we might need to solve is that in base we will have different version numbers so we somehow need to make this a variable.

But I am open for better ideas how to evolve the includes for new version of the spec.

PR #1782, opened 2023-08-02
fix typo in metaModel.json

glepnir:
typo fix . looks like this website not generate by a static tool ? main branch too old. so I fix in gh-page branch. If i am doing wrong branch please correct me thx :)


dbaeumer:
This got already addressed here: https://github.com/microsoft/vscode-languageserver-node/blob/f2ff7d55464a1f58f978cb6635bd8865f050553c/protocol/src/common/protocol.ts#L3411 from where we generate the meta model.

PR #1786, opened 2023-08-08
Add an explicit note about drive letter colons

DanTup:
See https://github.com/microsoft/vscode-languageserver-node/issues/1280


_includes/types/uri.md
```diff
@@ -16,6 +16,13 @@ We also maintain a node module to parse a string into `scheme`, `authority`, `pa
 
 Many of the interfaces contain fields that correspond to the URI of a document. For clarity, the type of such a field is declared as a `DocumentUri`. Over the wire, it will still be transferred as a string, but this guarantees that the contents of that string can be parsed as a valid URI.
 
+Care should be taken to handle encoded in URIs. Some clients (such as VS Code) may encode colons in drive letters while others do not. The following URIs should be considered equivalent and are both valid for clients and servers to use:
```
dbaeumer: I would actually phrase this more general saying that servers should handle all valid encoded URIs. For example clients might use either `file:///c:/project/readme.md` or `file:///c%3A/project/readme.md` to denote a file with a drive letter under Windows. However clients should be consistent in how they encode characters in URIs.

dbaeumer (changes requested):

_includes/types/uri.md
DanTup: > However clients should be consistent in how they encode characters in URIs.

I don't like the word _should_ 😁

I've pushed some tweaks - how's this? (I explicitly noted that one party should not assume the other party will use the same encoding, because I think it has to be that way - perhaps I should have noted the same about drive letter casing?)

rwols: What could also be problematic and where care should be taken is when clients send symlinked `file` URIs. Or dealing with case-insensitive filenames. I would argue a server cannot rely on the string-representation of `file` URIs.

DanTup: Yep, there are definitely more edge cases than this, but this one is simple (it's just encoding/decoding) and already caused me a problem. It's much less clear to me what the spec should say about symlinks or casing so I think they could be handled in their own PRs (assuming some rules can be agreed!).

DanTup:
@dbaeumer I pushed some tweaks based on your comments above. Would you like anything else changed?

_includes/types/uri.md
```diff
@@ -16,6 +16,13 @@ We also maintain a node module to parse a string into `scheme`, `authority`, `pa
 
 Many of the interfaces contain fields that correspond to the URI of a document. For clarity, the type of such a field is declared as a `DocumentUri`. Over the wire, it will still be transferred as a string, but this guarantees that the contents of that string can be parsed as a valid URI.
 
+Care should be taken to handle encoding in URIs. For example, some clients (such as VS Code) may encode colons in drive letters while others do not. The URIs below are both valid, but clients and servers should be consistent with the form they use themselves to ensure the other party doesn't interpret them as distinct URIs. Clients and servers should not assume that each other are encoding the same way (for example a client encoding colons in drive letters cannot assume server responses will have encoded colons).
+
+```
+file:///c:/project/readme.md
```
jwortmann, line 22: Maybe it would be better to use an uppercase drive letter in this example? Otherwise it could give the impression that drive letters must be lowercase, if both of the examples use a lowercase letter. Windows itself uses uppercase drive letters (almost) everywhere in the UI, so it seems more natural to use uppercase, and I remember bugs in the past with servers and with clients that expected a certain case for the drive letter.

dbaeumer: I agree with that suggestion @DanTup. May be best is to uppercase the drive letter only in one example.

dbaeumer (changes requested):

_includes/types/uri.md
DanTup: Done!

dbaeumer (approved):

aeschli (approved):

PR #1816, opened 2023-09-20
Update meta model

dbaeumer:


aeschli (approved):

wookayin:
I find that the 3.17 metaModel.json contains information for the succeeding version: `@since 3.18.0`. Is this intended?

https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.17/metaModel/metaModel.json

marcuscaisey:
>I find that the 3.17 metaModel.json contains information for the succeeding version: `@since 3.18.0`. Is this intended?
bump 🙏 

I'm generating types for my server using the 3.17 meta model and i'm not sure what i'm supposed to do with these proposed elements. I don't think that i should be generating code for them

dbaeumer:
Yes that is intended. Proposed features might change and they are implement in the vscode LSP client. If you don't need these leave them out. If you generate types for them make sure that users understand that they might change / break.

PR #1819, opened 2023-09-21
Add foldingRange kind for legal blurbs

mickaelistria:
That would allow clients who wants it to be informed that some regions are legal blurbs and collapse them by default.


mickaelistria:
Maybe `legal-comment` instead as the legal is usually 1 type of comment? This would allow clients that want to fold all comments to do so by folding all ranges of `kind.contains("comment")`.

dbaeumer:
@aeschli any take on that?

aeschli:
This doesn't fit well with the current kinds. We only support one kind per range. Legal is a subset of comment. We don't want to break clients that only know of `comment`, so we can't add `legal` to the default valueSet.
And how would a language server know that a comment is a 'legal' comment?
I think with will just make things confusing. I'm against that change.

dbaeumer:
@aeschli thanks for you comment.  I will close the PR then.

mickaelistria:
Thanks for the review. Particularly

> how would a language server know that a comment is a 'legal' comment?

I now agree with it, and I think it can indeed be better to place the logic in the client. most legal header are comments (you're fully right here) with special content (a SPDX header, a copyright notice...). The clients can take care of processing the comments for hints about whether the comment is just "legal" or not, fold it by default or not.

Issue #1461, opened 2022-05-02
Long diagnostic message field


ratmice:
In the bug I filed yesterday #1458 discusses tranfsering documents from server to client.
The impetus for that request though was dealing with long diagnostic messages.

Thinking through it a different solution than #1458 might be instead adding an `long_message?: string` to diagnostic.  This would appear to be a much simpler fix for the problem that I am dealing with, having less weirdness involving `uri` for the lsp server, and new request types.

robbycbennett:
For a long description, the LSP should support both dynamic and static descriptions. Here's an idea of how this could be implemented.

In `CodeDescription` add `dynamicDescription` as an optional string. Make `href` optional.
```ts
export interface CodeDescription {
	/**
	 * An URI to open with general information about the diagnostic error (e.g.
	 * paragraphs of a description, code examples, exceptions, links to sources).
	 * The URI can start with "file:", "https:", or other.
	 */
	href?: URI;

	/**
	 * A string with unique information about this particular instance of the
	 * diagnostic error (e.g. a warning that mentions your variable name
	 * specifcally and the function it was defined in).
	 */
	dynamicDescription?: string;
}
```

Add a `textDocument/publishDiagnosticLongDescriptions` method, as a notification from the server to the client. This is to avoid repetition of the long descriptions.

```ts
interface PublishDiagnosticLongDescriptionParams extends WorkDoneProgressParams {
	/**
	 * An array of long descriptions for each diagnostic
	 */
	diagnosticLongDescriptions: DiagnosticLongDescription[];
}
```

```ts
interface DiagnosticLongDescription {
	/**
	 * The diagnostic's code, which might appear in the user interface.
	 */
	code: integer | string;

	/**
	 * The static long description of a diagnostic (e.g. paragraphs of a
	 * description, code examples, exceptions, links to sources).
	 */
	longDescription: string;
}
```

robbycbennett:
This is a [highly requested feature](https://github.com/microsoft/vscode/issues/54272) with VS Code, specifically for long dynamic diagnostic descriptions.

PR #1825, opened 2023-10-12
add diagnostic dynamic long description

robbycbennett:
https://github.com/microsoft/language-server-protocol/issues/1461

### DIFF
--- a/_specifications/lsp/3.18/types/diagnostic.md
+++ b/_specifications/lsp/3.18/types/diagnostic.md
@@ -34,9 +34,21 @@ export interface Diagnostic {
 	source?: string;
 
 	/**
-	 * The diagnostic's message.
+	 * The diagnostic's message, e.g. 'Unknown type name "int_8_t"'.
 	 */
 	message: string;
+
+	/**
+	 * An optional property to describe this diagnostic's message with more
+	 * details. For example:
+	 * The type name "int_8_t" is unknown. Did you mean any of these types?
+	 * - int
+	 * - int8_t
+	 * Read more about the built-in types [here](file:///path/to/docs.md).
+	 *
+	 * @since 3.18.0
+	 */
+	longMessage?: string | MarkupContent;
 
 	/**
 	 * Additional metadata about the diagnostic.
--- a/_specifications/lsp/3.18/types/diagnostic.md
+++ b/_specifications/lsp/3.18/types/diagnostic.md
@@ -151,8 +151,19 @@ export interface DiagnosticRelatedInformation {
  */
 export interface CodeDescription {
 	/**
-	 * An URI to open with more information about the diagnostic error.
+	 * An URI to open with general information about the diagnostic error (e.g.
+	 * paragraphs of a description, code examples, exceptions, links to sources).
+	 * The URI can start with "file:", "https:", or other.
 	 */
-	href: URI;
+	href?: URI;
+
+	/**
+	 * A string with unique information about this particular instance of the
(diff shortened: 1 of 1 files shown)


robbycbennett:
@microsoft-github-policy-service agree company="Scientific Toolworks, Inc."

dbaeumer:
@RobbyCBennett tahnks for the PR. Changing the spec is the easy part :-). To contribute additional functionality to the spec a reference implementation is needed as well. See https://github.com/microsoft/language-server-protocol/blob/main/contributing.md

I will keep the PR open for a while but without a reference implementation I will not be able to merge the PR.

jwortmann:
Isn't this a breaking change?

Currently, the `href` field is mandatory. With this PR, it would become optional. So any client which unconditionally tries to read `href` might crash if the field doesn't exist.

robbycbennett:
> To contribute additional functionality to the spec a reference implementation is needed as well.

Thank you. I was just looking to get some feedback before the reference implementation is made.


> Isn't this a breaking change?

Yes, I will change it to accommodate for language clients that will crash after assuming the `href` field is defined.

Perhaps instead, it could be simplified to be part of `Diagnostic`, as a new field `longMessage` that could be just below `message`. I think that this would be more clear and it also wouldn't break language clients.

https://github.com/microsoft/language-server-protocol/pull/1825/commits/1348fc7c74f7427eb9376526725da673bd1e45ce

_specifications/lsp/3.18/types/diagnostic.md
```diff
@@ -34,10 +34,22 @@ export interface Diagnostic {
...
+	 * An optional property to describe this diagnostic's message with more
+	 * details. For example:
+	 * The type name "int_8_t" is unknown. Did you mean any of these types?
+	 * - int
+	 * - int8_t
+	 * Read more about the built-in types [here](file:///path/to/docs.md).
+	 *
+	 * @since 3.18.0
+	 */
+	longMessage?: string | MarkupContent;
```
rchl: The naming of the property can suggest that it's an alternative version of `message`. Not sure if the intention would be that clients show both or choose one.

If the intention is to show both then maybe this should be called `description`...

robbycbennett: Calling it `description` would be too similar to the property called `codeDescription`, which is a general/static description of the diagnostic.

In contrast, I thought about calling it `longMessage` because it is similar to `message` but provides more detail. Perhaps something like `messageDetails` would be more appropriate. I think that would infer that `message` is still important.

robbycbennett:
I didn't follow https://github.com/microsoft/language-server-protocol/blob/main/contributing.md because it seems to be intended for adding new types and methods.

Reference implementation:

Step 1: (this PR)
- Add messageDetails field to Diagnostic

Step 2: [PR in VSCode repo](https://github.com/microsoft/vscode/pull/195727):
- Add messageDetails field to Diagnostic (this is a different Diagnostic interface, defined in src/vscode-dts/index.d.ts)

Step 3
- Update the projects that get their 2 Diagnostic types from these 2 repos (like the npm module @types/vscode), which should prepare for step 4.

Step 4 ([PR in VSCode Language Server Node repo](https://github.com/microsoft/vscode-languageserver-node/pull/1339)):
- Change asDiagnostic function in createConverter and add asMessageDetails function (in client/src/common/codeConverter.ts)
- Change asDiagnostic function in createConverter and add asMessageDetails function (in client/src/common/protocolConverter.ts)

PR #1826, opened 2023-10-12
add method textDocument/publishDiagnosticLongDescriptions and add type DiagnosticLongDescription

robbycbennett:
https://github.com/microsoft/language-server-protocol/issues/1461


dbaeumer:
Same as https://github.com/microsoft/language-server-protocol/pull/1825

robbycbennett:
This could easily utilize a lot of memory. It is not easy to implement. I'm going to close it.

The existing URI functionality should be fine: files for languages servers for offline descriptions on the local machine and http/https requests for online descriptions.

PR #1831, opened 2023-10-20
Fix NotebookDocumentSync example

manandre:
The given example does not follow the specification just below.
The `NotebookSelector` field is an array type.


manandre:

dbaeumer (approved):

dbaeumer:
@manandre  Good catch. Thanks!

aeschli (approved):

Issue #1802, opened 2023-09-07
Support a "data"-like field on CompletionList that is also returned to the server in completionItem/resolve to avoid duplication in CompletionItem.data


DanTup:
(**Edit 2023-12-04**: This request has changed slightly throughout the discussion - see https://github.com/microsoft/language-server-protocol/issues/1802#issuecomment-1838387908 below)

---

(from https://github.com/microsoft/vscode-languageserver-node/pull/1237#issuecomment-1686687455)

There's an `itemDefaults.data` field for completion that allows `data` to be included once in a completion response rather than duplicated across all items.

For Dart, the `data` field contains a mix of data that is the same for all items (eg. the file the completion is being inserted into so we can compute edits for adding `import`s where required - since `/resolve` doesn't get any context) and data that is different (an ID to get back to the element being inserted so we can resolve things like documentation). Since `itemDefaults.data` replaces the whole of `data` we can't use it here, so we end up with a large payload with a lot of duplicated into.

It would be very helpful to have an option to merge `data` from items over the default (for ex. `Object.assign(itemDefaults.data, item.data)`?).

I'm happy to send PRs for this, but I want to agree an approach first:

- Do we support merge only for `data` or all fields?
- What should the merged default be called? (`itemDefaults.mergedData`?)
- Is `Object.assign(itemDefaults.data, item.data)` flexible enough? (you can use `null` to erase something from the defaults for a given item?)
- The capability can just be the client including the new field (eg. `mergedData`) in the `completionList.itemDefaults` set?

@dbaeumer WDYT?

puremourning:
Just remember that If we are going to specify a merge operation the client must perform, then the exact detailed merge behaviour and semantics must be fully specified by the protocol. 

Saying 'what JavaScript does' is not helpful if the client isn't JavaScript.

dbaeumer:
I do see the need for something like this due to performance reasons. But before we go down the pass can you collect some numbers that show how much this will speed things up (e.g time it take to show the completion in the UI). It will complicate the protocol and adds effort for clients to implement this and I want to ensure it is worth the effort.

DanTup:
[...]

PR #1834, opened 2023-10-25
Add a note about only calling completion/resolve for items in the last completion request

DanTup:
See https://github.com/microsoft/language-server-protocol/issues/1802#issuecomment-1778689197

@dbaeumer


dbaeumer:
I am closing the PR. See my comment here: https://github.com/microsoft/language-server-protocol/issues/1802#issuecomment-1805342424

PR #1862, opened 2023-12-14
Fixes #1644: Feature request: support tooltips for code lenses

### LINKED ISSUE #1644: Feature request: support tooltips for code lenses
In the VSCode extension API, a [Command](https://github.com/microsoft/vscode/blob/d4237bd8d548a45e2759f3b3fe4fb75dfeb677a9/src/vs/editor/common/languages.ts#L1455) can have an associated tooltip string. This allows the editor to display the tooltip when hovering over the code lens. Example:

image

The [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command) currently doesn't include this, making it difficult if not impossible to implement a code lens with a tooltip on the server side.

dbaeumer:
Fixes #1644

### DIFF
--- a/_specifications/lsp/3.18/types/command.md
+++ b/_specifications/lsp/3.18/types/command.md
@@ -8,10 +8,17 @@ interface Command {
 	 * Title of the command, like `save`.
 	 */
 	title: string;
+
+	/**
+	 * An optional tooltip.
+	 */
+	tooltip?: string;
+
 	/**
 	 * The identifier of the actual command handler.
 	 */
 	command: string;
+
 	/**
 	 * Arguments that the command handler should be
 	 * invoked with.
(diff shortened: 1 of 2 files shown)


_specifications/lsp/3.18/types/command.md
```diff
@@ -8,10 +8,17 @@ interface Command {
 	 * Title of the command, like `save`.
 	 */
 	title: string;
+
+	/**
+	 * An optional tooltip.
+	 */
+	tooltip?: string;
```
rchl, line 15: Does spec want to use a seemingly UI-centric "tooltip" wording? Would it be better to say, for example, "description" and let client choose to use it in the tooltip or elsewhere?

chrmarti (approved):

PR #1868, opened 2023-12-19
Clarify snippet escaping rules

DanTup:
The current text reads like you can escape $ and } where not strictly necessary, but according to VS Code's behaviour and https://github.com/microsoft/vscode/issues/201059 this is not the case - you may only escape the characters that are _required_ to be escaped, otherwise you'll see backslashes in the output.

VS Code Docs PR: https://github.com/microsoft/vscode-docs/pull/6928


dbaeumer (approved):

dbaeumer:
Thanks!

lramos15 (approved):

PR #1871, opened 2023-12-22
Mark Range in TextDocumentContentChangeEvent as optional

### LINKED ISSUE #1870: Range property in TextDocumentContentChangeEvent inconsistent with comment
In TextDocumentContentChangeEvent, the Range property isn't optional:
https://github.com/microsoft/language-server-protocol/blob/80b9d4afb6c60a5e19d8c2faee3e5b28cfb50735/_specifications/lsp/3.18/textDocument/didChange.md?plain=1#L74 

But the comment above do state that the event can be sent with only `text`:
```
/**
 * An event describing a change to a text document. If only a text is provided
 * it is considered to be the full content of the document.
 */
```

leitzler:
Fixes #1870

### DIFF
--- a/_specifications/lsp/3.18/textDocument/didChange.md
+++ b/_specifications/lsp/3.18/textDocument/didChange.md
@@ -71,7 +71,7 @@ export type TextDocumentContentChangeEvent = {
 	/**
 	 * The range of the document that changed.
 	 */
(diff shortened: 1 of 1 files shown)


dbaeumer:
See my comment in #1870

Issue #343, opened 2017-11-27
Collect new CompletionItemKinds


dbaeumer:

vladdu:
- pragma / directive

vladdu:
- macro

rictic:
Slot, in the sense of a named, declared hole in a template that a user of the element can use to insert content.

e.g. `<slot>` in Shadow Dom v1, `<content>` in Shadow Dom v0, transclusion points in AngularJS, `<ng-content>` in Angular.

laughedelic:
There is `SymbolKind`, but no `CompletionItemKind` for  `Namespace`

smarter:
There's SymbolKind.Package, but not CompletionItemKind.Package

kdvolder:
This is futile, server should be able to define their own completion kinds. There's no way you are going to be able to enumerate all possible completion kinds in the world. Especially if you are defining a language server for a dsl it seems unreasonable to bake your domain terminology into the LSP spec. (The item kinds would be specific to the language domain and could be things like , 'task', 'job', 'pipeline', 'resource').

kdvolder:
Or... server must be able to provide icons somehow. But yes, I understand how that may not naturally blend with the look and feel of every client.

I suppose having a growing list of 'standard' kinds is still helpful. Probably doesn't help with DSL languages but at least there's a larger set of 'good looking icons' to pick from. Maybe I was a bit harsh calling this 'futile'. So please do carry on enlarging the list :-)

dbaeumer:
Will do :-)

@jrieken do you think `marco` and `directive/pragma` is something we would like to VS Code as well. They are for sure useful for C/C++

jrieken:
> Or... server must be able to provide icons somehow. But yes, I understand how that may not naturally blend with the look and feel of every client.

That. The fear is that it will look horrible because every editor has a different style. On other hand we also don't want editor authors to have to provide 100s of icons. 

> @jrieken do you think marco and directive/pragma is something we would like to VS Code as well. They are for sure useful for C/C++

In VS Code there is a shared subset approach and the rule of thumb is that there are no 1-offs for one language but that you have name five other languages that have the concept.

dbaeumer:
Fair point. The language I know that have these concepts are:

- C/C++
- Lisp
- C# (has #ifdef, but not macros)
- Assembler languages :-)

kdvolder:
[...]

PR #1872, opened 2023-12-26
Add Instruction and Register variants to CompletionItemKind

WillLillis:
Would it be possible to add `Instruction` and `Register` variants to `CompletionItemKind` to better support LSPs targeting assembly languages?


dbaeumer:
See #343.

Usualy simply adding them to the spec without a client supporting them will not gain a lot.

PR #1892, opened 2024-02-01
Add snippet text edit specification

### LINKED ISSUE #724: Allow CodeActions to specify cursor position
EDIT: We now documented this as a proper extension: https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#snippet-textedit. lsp-mode for emacs implements it: emacs-lsp/lsp-mode@9eb3324

`WorkspaceEdit` intentionally avoids speaking about cursor position, just like the rest of the protocol. 

We however slightly bend this rule for completions, where we allow `$0` snippets which is a way to place the cursor. 

I want to argue that a similar feature is required for code actions.

Here are a couple of examples:

**add derive**

For `add derive` in rust-analyzer, I want to turn this (`|` is the cursor)

```rust
struct Foo {
    f: f64,|
}
```

into this

```rust
#[derive(|)]
struct Foo {
    f: f64,
}
```

It's pretty important that `|` is in the derive, such that the user can immediately type/complete `#[derive(Debug, Clone, Copy)]` incantation. Furthermore, the code action is available even if `#[derive]` already exists on the type: in this case, *all* the action does is it moves the cursor (which is still very convenient for the user)

**create mod**

Again, in rust-analyzer, for `mod foo;` I want to provide a fix which creates an `foo.rs` file, opens it in the editor and places the cursor at the start of the file (I can imagine an extension where the file is pre-filled with licensing boilerplate, and the cursor should be after it). Note that this is similar to https://github.com/Microsoft/language-server-protocol/issues/612

Note that, unlike the completions case, we can't just special-case code actions and add `insertTextFormat` on code-action itself: the reason is that lazy code-actions are resolved in a round-about way via a command and `applyWorkspaceEdit` request sent from the server to the client. So looks like we need to add this ability directly to `WorkspaceEdit`. Perhaps adding an

```
insertTextFormat?: InsertTextFormat
```

to the `TextEdit` (with a corresponding opt-in capability) will do the trick? The `insertTextFormat` on `CompletionItem` then becomes redundant though :)

MariaSolOs:
Follow-up of https://github.com/microsoft/vscode-languageserver-node/pull/1343.
This time it actually closes https://github.com/microsoft/language-server-protocol/issues/724.

### DIFF
--- a/_specifications/lsp/3.18/types/textEdit.md
+++ b/_specifications/lsp/3.18/types/textEdit.md
@@ -84,6 +84,10 @@ export interface AnnotatedTextEdit extends TextEdit {
 
 Since 3.18.0, there is also the concept of a snippet text edit, which supports inserting a snippet instead of plain text.
 
+Some important remarks:
+- For each file, only one snippet can specify a cursor position. In case there are multiple snippets defining a cursor position for a given URI, it is up to the client to decide the end position of the cursor.
+- In case the snippet text edit corresponds to a file that is not currently open in the editor, the client should downgrade the snippet to a non-interactive normal text edit and apply it to the file. This ensures that a workspace edit doesn't open arbitrary files.
+
 <div class="anchorHolder"><a href="#snippetTextEdit" name="snippetTextEdit" class="linkableAnchor"></a></div>
 
 ```typescript
--- a/_specifications/lsp/3.18/types/textEdit.md
+++ b/_specifications/lsp/3.18/types/textEdit.md
@@ -1,6 +1,8 @@
-#### <a href="#textEdit" name="textEdit" class="anchor"> TextEdit & AnnotatedTextEdit </a>
+#### <a href="#textEdit" name="textEdit" class="anchor"> TextEdit, AnnotatedTextEdit & SnippetTextEdit </a>
 
-> New in version 3.16: Support for `AnnotatedTextEdit`.
+- New in version 3.16: Support for `AnnotatedTextEdit`.
+- New in version 3.18: Support for `SnippetTextEdit`.
+
 
 A textual edit applicable to a text document.
 
@@ -77,5 +79,33 @@ export interface AnnotatedTextEdit extends TextEdit {
 	 * The actual annotation identifier.
 	 */
 	annotationId: ChangeAnnotationIdentifier;
+}
+```
+
+Since 3.18.0, there is also the concept of a snippet text edit, which supports inserting a snippet instead of plain text.
+
+<div class="anchorHolder"><a href="#snippetTextEdit" name="snippetTextEdit" class="linkableAnchor"></a></div>
+
+```typescript
+/**
+ * An interactive text edit.
+ *
+ * @since 3.18.0
+ */
+export interface SnippetTextEdit {
+	/**
+	 * The range of the text document to be manipulated.
+	 */
+	range: Range;
+
+	/**
+	 * The snippet to be inserted.
+	 */
+	snippet: StringValue;
+
+	/**
+	 * The actual identifier of the snippet edit.
+	 */
+	annotationId?: ChangeAnnotationIdentifier;
 }
 ```
--- a/_data/linkableTypes.yml
+++ b/_data/linkableTypes.yml
@@ -900,3 +900,5 @@
     link: '#inlineCompletionItem'
   - type: 'StringValue'
     link: '#stringValue'
+  - type: 'SnippetTextEdit'
+    link: '#snippetTextEdit'
--- a/_specifications/lsp/3.18/types/textDocumentEdit.md
+++ b/_specifications/lsp/3.18/types/textDocumentEdit.md
@@ -1,6 +1,8 @@
 #### <a href="#textDocumentEdit" name="textDocumentEdit" class="anchor"> TextDocumentEdit </a>
 
-> New in version 3.16: support for `AnnotatedTextEdit`. The support is guarded by the client capability `workspace.workspaceEdit.changeAnnotationSupport`. If a client doesn't signal the capability, servers shouldn't send `AnnotatedTextEdit` literals back to the client.
+* New in version 3.16: support for `AnnotatedTextEdit`. The support is guarded by the client capability `workspace.workspaceEdit.changeAnnotationSupport`. If a client doesn't signal the capability, servers shouldn't send `AnnotatedTextEdit` literals back to the client.
+
+* New in version 3.18: support for `SnippetTextEdit`. The support is guarded by the client capability `workspace.workspaceEdit.snippetEditSupport`. If a client doesn't signal the capability, servers shouldn't send `SnippetTextEdit` snippets back to the client.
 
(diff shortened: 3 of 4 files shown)


_specifications/lsp/3.18/types/textEdit.md
```diff
@@ -79,3 +81,31 @@ export interface AnnotatedTextEdit extends TextEdit {
...
+export interface SnippetTextEdit {
+	/**
+	 * The range of the text document to be manipulated.
+	 */
+	range: Range;
+
+	/**
+	 * The snippet to be inserted.
+	 */
+	snippet: StringValue;
```
MariaSolOs, line 108: @dbaeumer It seems like `StringValue`s haven't been documented yet, despite already being used in the metamodel for `InlineValue`s. Should we add that specification in this PR as well?

rchl:
Can you also check out my questions in https://github.com/microsoft/language-server-protocol/issues/724#issuecomment-1895943206 and potentially address in the spec?

_specifications/lsp/3.18/types/textEdit.md
```diff
@@ -79,3 +81,31 @@ export interface AnnotatedTextEdit extends TextEdit {
 	annotationId: ChangeAnnotationIdentifier;
 }
 ```
+
+Since 3.18.0, there is also the concept of a snippet text edit, which supports inserting a snippet instead of plain text.
```
mfussenegger, line 85: Shouldn't this have some client capability associated, so servers know if they can make use of it?

_specifications/lsp/3.18/types/textEdit.md
dbaeumer: Good point. I would add it to this PR.

dbaeumer:
@rchl I commented here: https://github.com/microsoft/language-server-protocol/issues/724#issuecomment-1921011006

_specifications/lsp/3.18/types/textEdit.md
dbaeumer: There is one ion the proposed implementation so it should definitely be in the spec.

jwortmann:
To make snippets with multiple tab stops (https://github.com/microsoft/language-server-protocol/issues/724#issuecomment-483206386) work properly, clients likely need to use built-in editor functionality to insert the text as a snippet. This might imply that the editor automatically re-indents inserted text which contains multiple lines; for example it seems to be the case in VSCode (https://github.com/microsoft/language-server-protocol/issues/724#issuecomment-631558796), and it also happens in Sublime Text. A similar situation for when snippets are inserted as part of autocompletions is solved by the client announcing the [`insertTextMode`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertTextMode) and `insertTextModeSupport` properties in its completion capability. Will this be handled for SnippetTextEdit in a similar manner, so that servers know how to format the whitespace in multi-line snippets?

An alternative would be to declare in the specs that SnippetTextEdits are only allowed to contain a single tab stop. Then clients could do a raw insert of the text and after that manually modify the cursor position in a second step.

_specifications/lsp/3.18/types/textEdit.md
MariaSolOs: Absolutely, the reason this is a draft PR is because it's incomplete :) I just published it to ask about the `StringValue` thing.

_specifications/lsp/3.18/types/textEdit.md
MariaSolOs: I decided to create a separate PR: https://github.com/microsoft/language-server-protocol/pull/1893 ;)

MariaSolOs:
@dbaeumer Should I make any changes to the metamodel, or is it done automatically?

MariaSolOs:
@dbaeumer with https://github.com/microsoft/vscode-languageserver-node/issues/1414 fixed, I think this PR is ready :)

dbaeumer:
I think we should add some additional clarification to the spec about some expectations. Things I have in mind:

- if there are more than one snippet for a file which defines a cursor position then the end position is determined by the client.
- if a snippet is specified for the non active editor the client should downgrade the snippet to a normal text edit and apply it to the file. This avoids that a workspace edit when applied opens up random files.

Since snippets it workspace edits are most likely used with code actions there will be an open active editor to which snippets can be applied.

@MariaSolOs can you add this to the PR?

MariaSolOs:
> * if a snippet is specified for the non active editor the client should downgrade the snippet to a normal text edit and apply it to the file. This avoids that a workspace edit when applied opens up random files.

By "apply it to the file", do you mean the currently open editor, or the closed one specified by the snippet edit?

If it's the currently opened editor, that sounds counterintuitive since the server (and user) expect the change to happen in the specified editor.

If it's the non-active editor, how does that prevent the workspace from opening up random files?

dbaeumer:
> By "apply it to the file", do you mean the currently open editor, or the closed one specified by the snippet edit?

No to the file that is specific to the snippet edit without opening the file in the editor. This is what is done for normal text edits as well.

dbaeumer (approved):

jrieken (approved):

jwortmann:
Just to confirm, because there wasn't any response to my question from https://github.com/microsoft/language-server-protocol/pull/1892#issuecomment-1921520641, and it's unclear from the text in the specs:

Is it correct to assume that language servers have to figure out by themselves (for examply using the `clientInfo` property from the initialize request) what kind of indentation (InsertTextMode) the client expects for SnippetTextEdits with multiple tab stops?

dbaeumer:
@jwortmann sorry for that. The `InsertTextMode` is only valid for completion items. Workspace edits in general don't have support for automatic indentation. So they are always `asIs`.

jwortmann:
> @jwortmann sorry for that. The `InsertTextMode` is only valid for completion items. Workspace edits in general don't have support for automatic indentation. So they are always `asIs`.

The problem with `asIs` is, that it is simply not possible to imlement at the client side. At least not for clients which are implemented in form of a third-party plugin, relying on the API and built-in functionalities from the editor, and in case the snippet functionality automatically adds the whitespace from the current indentation level to the whitespace provided in the snippet content. For example this happens in Sublime Text, where snippets can be inserted via a built-in command like
```python
view.run_command('insert_snippet', {'contents': edit['snippet']['value']})
```
and without any means to change indentation behavior. I wonder how does VS Code solve this problem, because the comment I linked above (https://github.com/microsoft/language-server-protocol/issues/724#issuecomment-631558796) indicates that the same happens in VS Code. Also compare https://github.com/microsoft/language-server-protocol/issues/83#issuecomment-269763139 explaining the "magic indentation" that happens in VS Code. As far as I understand, `insertTextMode` was introduced to make multi-line completions work properly in clients which have no direct control about this. Now the same problem will happen with SnippetTextEdits if they contain multiple lines.

I described a possible workaround in my comment above, which should be sufficient if setting the final cursor position is the only goal of the SnippetTextEdit:
> An alternative would be to declare in the specs that SnippetTextEdits are only allowed to contain a single tab stop. Then clients could do a raw insert of the text and after that manually modify the cursor position in a second step.

But in that case this should be explained in the specs. Because currently the snippet content only has a link to the [StringValue](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#stringValue), and its description explicitly mentions multiple tab stops.

PR #1893, opened 2024-02-01
Add string value definition

MariaSolOs:
Adds the definition of `StringValue`, which is currently used in the metamodel by inline completions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ npm-debug.log
 _site/
 .bundle/
 vendor/
+.jekyll-metadata
--- a/_data/linkableTypes.yml
+++ b/_data/linkableTypes.yml
@@ -898,3 +898,5 @@
     link: '#inlineCompletionList'
   - type: 'InlineCompletionItem'
     link: '#inlineCompletionItem'
+  - type: 'StringValue'
+    link: '#stringValue'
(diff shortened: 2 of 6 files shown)


MariaSolOs:
@dbaeumer I think we should remove the `insertTextFormat` property from `InlineCompletionItem`:
https://github.com/microsoft/language-server-protocol/blob/a1766bb2c70bb8e847a22ca99f933047716d9a64/_specifications/lsp/3.18/language/inlineCompletion.md?plain=1#L233-L237

Why?
- It becomes redundant with the type of `insertText` being `string | StringValue`.
- [The metamodel doesn't have such property](https://github.com/microsoft/language-server-protocol/blob/b9c44297035bbad48215e6459ef8051c1f225ae0/_specifications/lsp/3.18/metaModel/metaModel.json#L4170).
- [The analogous VS Code definition doesn't have such property either](https://github.com/microsoft/vscode-languageserver-node/blob/295aaa393fda8ecce110c38880a00466b9320e63/types/src/main.ts#L4376)

dbaeumer (approved):

alexr00 (approved):

dbaeumer:
@MariaSolOs makes perfekt sense to remove `insertTextFormat`

MariaSolOs:
> @MariaSolOs makes perfekt sense to remove `insertTextFormat`

Alright, doing that here: #1894 :)

PR #1905, opened 2024-03-01
Specification for MarkupContent support in diagnostic messages

### LINKED ISSUE #250: Change the type of Diagnostic.message to MarkedString
Currently, diagnostic messages do not carry any formatting information, this forces clients to be conservative when displaying them, see https://github.com/Microsoft/vscode/issues/22239 and the proposed solution:
> Going forward. Deprecate the current way diagnostics are rendered and make them work just like the regular hover: a Markdown string which defaults to a non-monospace font but still lets language feature providers open up monospace code blocks with triple back quotes.

MariaSolOs:
Closes https://github.com/microsoft/language-server-protocol/issues/250 by introducing a new client capability in `workspace.diagnostic.markupMessageSupport`, which when set allows language server to send diagnostic messages as `MarkupContent`.

Note that several language servers today (like LuaLS or rust-analyzer) already return diagnostics with Markdown syntax (such as backticks). For a better UI experience, formalizing this in the spec is desired.

The implementation was done in Neovim [here](https://github.com/neovim/neovim/pull/27693).

### DIFF
--- a/_specifications/lsp/3.18/types/diagnostic.md
+++ b/_specifications/lsp/3.18/types/diagnostic.md
@@ -1,5 +1,9 @@
 #### <a href="#diagnostic" name="diagnostic" class="anchor"> Diagnostic </a>
 
+- New in version 3.18: support for markup content in diagnostic messages. The support is guarded by the 
+client capability `textDocument.diagnostic.markupMessageSupport`. If a client doesn't signal the capability, 
+servers shouldn't send `MarkupContent` diagnostic messages back to the client.
+
 Represents a diagnostic, such as a compiler error or warning. Diagnostic objects are only valid in the scope of a resource.
 
 ```typescript
@@ -35,8 +39,11 @@ export interface Diagnostic {
 
 	/**
 	 * The diagnostic's message.
+	 * 
+	 * @since 3.18.0 - support for MarkupContent. This is guarded by the client
+	 * capability `textDocument.diagnostic.markupMessageSupport`.
 	 */
-	message: string;
+	message: string | MarkupContent;
 
 	/**
 	 * Additional metadata about the diagnostic.
--- a/_specifications/lsp/3.18/general/initialize.md
+++ b/_specifications/lsp/3.18/general/initialize.md
@@ -886,6 +886,28 @@ interface ServerCapabilities {
 	inlineCompletionProvider?: boolean | InlineCompletionOptions;
 
 	/**
+	 * Text document specific server capabilities.
+	 * 
+	 * @since 3.18.0
+	 */
+	textDocument?: {
+		/**
+		 * Capabilities specific to the diagnostic pull model.
+		 * 
+		 * @since 3.18.0
+		 */
+		diagnostic?: {
+			/**
+			 * Whether the server supports `MarkupContent` in diagnostic messages.
+			 * 
+			 * @since 3.18.0
+			 * @proposed
+			 */
+			markupMessageSupport?: boolean;
+		};
+	};
+
+	/**
 	 * Workspace specific server capabilities
 	 */
 	workspace?: {
--- a/_specifications/lsp/3.18/language/codeAction.md
+++ b/_specifications/lsp/3.18/language/codeAction.md
@@ -361,6 +361,9 @@ export interface CodeActionContext {
 	 * for the given range. There is no guarantee that these accurately reflect
 	 * the error state of the resource. The primary parameter
 	 * to compute code actions is the provided range.
+	 * 
+	 * Note that the client should check the `textDocument.diagnostic.markupMessageSupport`
+	 * server capability before sending diagnostics with markup messages to a server.
 	 */
 	diagnostics: Diagnostic[];
 
--- a/_specifications/lsp/3.18/language/pullDiagnostics.md
+++ b/_specifications/lsp/3.18/language/pullDiagnostics.md
@@ -30,6 +30,14 @@ export interface DiagnosticClientCapabilities {
 	 * pulls.
 	 */
 	relatedDocumentSupport?: boolean;
+
+	/**
+	 * Whether the client supports `MarkupContent` in diagnostic messages.
+	 * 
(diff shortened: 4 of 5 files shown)


_specifications/lsp/3.18/types/diagnostic.md
```diff
@@ -35,8 +39,11 @@ export interface Diagnostic {
 
 	/**
 	 * The diagnostic's message.
+	 * 
+	 * @since 3.18.0 - support for MarkupContent. This is guarded by the client
+	 * capability `workspace.diagnostic.markupMessageSupport`.
 	 */
-	message: string;
+	message: string | MarkupContent;
```
rchl, line 46: It will be a breaking change (type-wise) if the type suddenly changes from `string` to `string | MarkupContent`. I think this might need to be a separate property.

(not sure how such cases were handled historically)

rchl: Realistically servers will probably have to support both plain string and markup since there is no guarantee that all clients will support markup. So it might make sense for servers to provide both variants in separate properties (when client supports markup).

MariaSolOs: > It will be a breaking change (type-wise) if the type suddenly changes from string to string | MarkupContent. I think this might need to be a separate property.

I think this is fine given that it is behind a capability. Note how previous proposals followed the same approach; e.g. [snipped edits](https://github.com/microsoft/language-server-protocol/pull/1892).

MariaSolOs: > Realistically servers will probably have to support both plain string and markup since there is no guarantee that all clients will support markup. So it might make sense for servers to provide both variants in separate properties (when client supports markup).

True, although one could argue that this is also true for other features that support markup content, yet those don't provide both formats.

MariaSolOs: That being said, @dbaeumer has the final call.

MariaSolOs:
Note to self: Update the meta model (unless it's autogenerated?)

dbaeumer (commented):
> That being said, @dbaeumer has the final call. 

If objects only flow between the same client and server a capability is enough on the receiving side. However diagnostic can be more tricky since the flow in code actions from the client to the server. And in principal diagnostics from a plugin A can flow to a a server B. So we would need a server capability as well to let clients know that the can handle diagnostics with markup content.

MariaSolOs:
> > That being said, @dbaeumer has the final call.
> 
> If objects only flow between the same client and server a capability is enough on the receiving side. However diagnostic can be more tricky since the flow in code actions from the client to the server. And in principal diagnostics from a plugin A can flow to a a server B. So we would need a server capability as well to let clients know that the can handle diagnostics with markup content.

Makes sense. I'll add the server capability too.

I am unsure of where exactly this capability should be defined though. I don't believe we should add it under `diagnosticsProvider`, since a non-provider could still receive diagnostics. Do you think we should add it at the top level @dbaeumer?

dbaeumer:
I would mirror what we have on the client

```
ServerCapabilities {
    textDocument: {
         diagnostics: {
         }
    }
}
```

MariaSolOs:
> I would mirror what we have on the client
> 
> ```
> ServerCapabilities {
>     textDocument: {
>          diagnostics: {
>          }
>     }
> }
> ```

I assume you're referring to `workspace` capabilities ;)

EDIT: Oh wait, I just realized that there's `workspace.diagnostics` and `textDocument.diagnostic` capabilities. I had initially added to the former one, but it makes more sense to add it to text document capabilities.

MariaSolOs:
> > That being said, @dbaeumer has the final call.
> 
> If objects only flow between the same client and server a capability is enough on the receiving side. However diagnostic can be more tricky since the flow in code actions from the client to the server. And in principal diagnostics from a plugin A can flow to a a server B. So we would need a server capability as well to let clients know that the can handle diagnostics with markup content.

@dbaeumer I've addressed this in https://github.com/microsoft/language-server-protocol/commit/406cd2357f23f1fe9c6b66315759a252770b2a5c :)

dbaeumer (commented):
@MariaSolOs  PR looks good to me. Would you be willing to make a PR against the vscode-languageserver-node repository as well. We would need it for the meta model and we should set the capabilities to false for VS Code.

mfussenegger (commented):
What should a client do if the server sends diagnostics with markup, but has `markupMessageSupport` set to false?

Would a client then need to filter out those diagnostics in places where the client sends them back to the server (e.g. code actions)?

MariaSolOs:
> @MariaSolOs PR looks good to me. Would you be willing to make a PR against the vscode-languageserver-node repository as well. We would need it for the meta model and we should set the capabilities to false for VS Code.

@dbaeumer sure, but now that we have an editor implementation and the meta model has been updated here as well, could we merge this first?

dbaeumer:
@MariaSolOs just out of curiosity: how did you update the meta model?

dbaeumer (approved):

bpasero (approved):

MariaSolOs:
> @MariaSolOs just out of curiosity: how did you update the meta model?

Ehhh I did it manually, with lots of care and love hehe

mfussenegger:
> What should a client do if the server sends diagnostics with markup, but has `markupMessageSupport` set to false?
> 
> Would a client then need to filter out those diagnostics in places where the client sends them back to the server (e.g. code actions)?

Any chance to have a clarification for this in the specification? The PR against neovim currently converts the MarkupContent to a string by simply taking the value from MarkupContent unmodified, dropping the kind information.

Other possible options I see:

- Exclude diagnostics with MarkupContent if the server has no support for it
- Ignore the server capability but ensure only diagnostics originating from a server are sent to it on code actions requests (but then why require the extra server capability?)

dbaeumer:
> Any chance to have a clarification for this in the specification? 

Agree we should do that. And we should exclude those diagnostics from the servers that don't have support for them.

PR #1966, opened 2024-07-04
Clarify deltaLine and deltaStart of SemanticTokens payload

### LINKED ISSUE #1855: Clarify docs: deltaLine and deltaStart fields of semantic tokens relate to the _start_ of the previous token
In writing an LSP-based language server, I've recently fumbled for a while assuming that these fields related to the _end_ of any previous token. The highlighting _just wasn't working_ the way I expected to, because of this assumption.

I now understand why the format is the way it is - to support overlapping tokens, in the languages those are relevant, but this wasn't obvious to me.

I'd be happy to update relevant pieces of docs via PRs.

I know of 2 relevant places to update:
- [x] in these lines of the spec:
  ```
  at index 5*i - deltaLine: token line number, relative to the previous token
  at index 5*i+1 - deltaStart: token start character, relative to the previous token (relative to 0 or the previous token’s start if they are on the same line)
  ```
  
  (simply adding wording like "relative to **the start of** the previous token...")

- [ ] ~~the de facto Node sample~~

any other places that would be useful to clarify in?

StachuDotNet:
Resolves #1855


StachuDotNet:

dbaeumer (approved):

aeschli (approved):

PR #1993, opened 2024-08-05
Update ruby version, add GlobPatterns to document filters for 3.18 and update meta model to latest implementation.

dbaeumer:


bpasero (approved):

DanTup:
@dbaeumer is it a breaking change to change `pattern` from `string` to `GlobPattern`? I have code that is assuming `pattern` is a string and when generating code against 3.18 it seems it could be a `RelativePattern`. While I can update this, I think passing `RelativePattern`s in places that were previously strings may need a capability to avoid the potential for breaking already-shipped clients/servers by passing `RelativePatterns`?

dbaeumer:
Good catch. We need to have a client capability so that servers don't send `RelativePatterns` to clients which only expect strings.

PR #2001, opened 2024-08-12
Added Draco.Lsp as an LSP SDK in .NET

LPeter1997:
We have our own LSP implementation that a few people have already found simple and useful. Available as a NuGet package [here](https://www.nuget.org/packages/Draco.Lsp/0.3.6-pre).


LPeter1997:

dbaeumer (approved):

bpasero (approved):

Issue #2011, opened 2024-09-02
Why signature help doesn't provide the version of current document?


imbant:
In the [document](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelpParams), the only thing textDocument/signatureHelp provides about the current document is the [uri](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentIdentifier) of it.

If the language server don't know the version, is it possible to confirm or constrain the didChange request to be later than signatureHelp ?  Because usually when the document content changes, the language server needs to re-compile to obtain the information required by signature help. The language server needs to handle signature help after compiling, or, the language server can know that the content didn't change by comparing document versions between signatureHelp and didChange

dbaeumer:
The LSP protocol ensures that any outstanding document change requests are sent to the server. The server should see the document in exactly the same state as the client saw it when the signature help request got triggered. If you see a different behavior please open a corresponding issue against the client implementation. For VS Code here: https://github.com/Microsoft/vscode-languageserver-node

imbant:
@dbaeumer  Hi, thank you for your attention.

I think there is a same issue for both client and server:
Is [textDocument/didChange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didChange) always before [textDocument/signatureHelp](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp) ?
It seems LSP doesn't declare it.

If not, for my case, `signatureHelp` may need the textdocument's version to make sure `didChange` happened and can get access to the latest semantic information, because my language server will recompile after `didChange`, rather than `signatureHelp`

dbaeumer:
I am not sure I understand this. The spec is clear that clients can't reorder events and that servers should only reorder events if they know what they do. Can you point me to an example were you receive a request before the change event is received and it should actually have happened the other way around?

If you need the document version it is part of the change event. So you server can ensure that the recompiled version adheres to the version number when receving signature help.

imbant:
I think the way I expressed it before was misleading. Let me rephrase the question: After the user hits the keyboard, can the language server confirm that the change event always occurs before the signature help event?

I debugged it and it seems to be the case, but the protocol does not seem to have this requirement. So I want to confirm

dbaeumer:
[...]

PR #2017, opened 2024-09-11
Ensure document state synchronization before client requests

imbant:
Copy this [discussion](https://github.com/microsoft/language-server-protocol/issues/2011#issuecomment-2343081904) to the document.

This content could help language server developers to safely handle signature help and completion requests after the compilation and I think the compilation is usually triggered by `didChange` event.

### DIFF
--- a/_specifications/lsp/3.17/textDocument/didChange.md
+++ b/_specifications/lsp/3.17/textDocument/didChange.md
@@ -4,6 +4,14 @@ The document change notification is sent from the client to the server to signal
 
 Before requesting information from the server (e.g., `textDocument/completion` or `textDocument/signatureHelp`), the client must ensure that the document's state is synchronized with the server to guarantee reliable results.
 
+The following example shows how the client should synchronize the state when the user has continuous input, assuming user input triggered `textDocument/completion`:
+
+| Document Version | User Input          | Client Behavior                                 | Request                   |
+| ---------------- | ------------------- | ----------------------------------------------- | ------------------------- |
+| 5                | document change one | sync document `v5` to the server                | `textDocument/didChange`  |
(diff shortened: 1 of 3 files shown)


_specifications/lsp/3.17/textDocument/didChange.md
```diff
@@ -2,6 +2,8 @@
 
 The document change notification is sent from the client to the server to signal changes to a text document. Before a client can change a text document it must claim ownership of its content using the `textDocument/didOpen` notification. In 2.0 the shape of the params has changed to include proper version numbers.
 
+Before requesting information from the server (e.g., `textDocument/completion` or `textDocument/signatureHelp`), the client must ensure that the document's state is synchronized with the server to guarantee reliable results.
```
dbaeumer, line 5: I think we need to make this more concrete since this sentences allows for interpretation as well (e.g. another change occurers before the client sends the request). So in a sequence

`<document change one><code complete><document change two>` the client needs to sync `<document change one>` to the server before requesting code compltion but not document change two.

dbaeumer (changes requested):
See comment

_specifications/lsp/3.17/textDocument/didChange.md
imbant: Sure. I added your example with a timeline table

imbant:
> See comment

Hi, is there anything I can improve on in my PR?

Bunggie (approved):

dbaeumer (approved):

lszomoru (approved):

imbant:
Wow thanks for your great work, as a language server developer it's my pleasure to contribute to this protocol :)

PR #2018, opened 2024-09-11
Add support for `completionList.applyKind` to determine how values from `completionList.itemDefaults` and `completion` are combined.

### LINKED ISSUE #1802: Support a "data"-like field on CompletionList that is also returned to the server in completionItem/resolve to avoid duplication in CompletionItem.data

DanTup:
The goal here is to significantly reduce payload sizes by allowing the server to avoid duplicating values across all `completion.commitCharacters` and `completion.data` that are the same for all (or most) items.

Only the fields specified in the spec support `merge` because the exact merge semantics must be specified. A client capability indicates which fields a given client supports merging.

@dbaeumer I implemented it this way because I worry that a single boolean may cause issues in future if new fields are added that we might want to support merge for, but I am happy to change if you don't like this.

Fixes https://github.com/microsoft/language-server-protocol/issues/1802

### DIFF
--- a/_specifications/lsp/3.18/language/completion.md
+++ b/_specifications/lsp/3.18/language/completion.md
@@ -166,6 +166,18 @@ export interface CompletionClientCapabilities {
 		 * @since 3.17.0
 		 */
 		itemDefaults?: string[];
+
+		/**
+		 * Specifies which fields of `CompletionList.applyKind` the client
+		 * supports. If omitted, no properties are supported and all fields
+		 * in a completion item will replace the defaults.
+		 * 
+		 * Clients may only specify fields that have merge rules defined in the
+		 * LSP spec.
+		 *
+		 * @since 3.18.0
+		 */
+		applyKinds?: string[];
 	}
 }
 ```
@@ -338,7 +350,8 @@ export interface CompletionList {
 	 * be used if a completion item itself doesn't specify the value.
 	 *
 	 * If a completion list specifies a default value and a completion item
-	 * also specifies a corresponding value, the one from the item is used.
+	 * also specifies a corresponding value, the rules for combining these are
+	 * defined by `applyKinds`, defaulting to "replace".
 	 *
 	 * Servers are only allowed to return default values if the client
 	 * signals support for this via the `completionList.itemDefaults`
@@ -384,6 +397,68 @@ export interface CompletionList {
 		 * @since 3.17.0
 		 */
 		data?: LSPAny;
+	}
+
+	/**
+	 * Specifies how fields from a completion item should be combined with those
+	 * from `completionList.itemDefaults`.
+	 * 
+	 * In unspecified, all fields will be treated as "replace".
+	 * 
+	 * If a field's value is "replace", the value from a completion item will
+	 * always be used instead of the value from `completionItem.itemDefaults`.
+	 * 
+	 * If a field's value is "merge", the values will be merged using the rules
+	 * defined against each field below.
+	 *
+	 * Servers are only allowed to return `applyKind` if the client
+	 * signals support for this via the `completionList.applyKinds`
+	 * capability.
+	 *
+	 * @since 3.18.0
+	 */
+	applyKind?: {
+		/**
+		 * Specifies whether commitCharacters on a completion will replace or be
+		 * merged with those in `completionList.itemDefaults.commitCharacters`.
+		 * 
+		 * If "replace", the commit characters from the completion item will
+		 * always be used unless not provided, in which case those from
+		 * `completionList.itemDefaults.commitCharacters` will be used. An empty
+		 * list can be used if a completion item does not have any commit
+		 * characters and also should not use those from
+		 * `completionList.itemDefaults.commitCharacters`.
+		 * 
+		 * If "merge" the commitCharacters for the completion will be the union
+		 * of all values in both `completionList.itemDefaults.commitCharacters`
+		 * and the completion's own `commitCharacters`.
+		 *
+		 * @since 3.18.0
+		 */
+		commitCharacters?: "replace" | "merge";
+
+		/**
+		 * Specifies whether data on a completion will replace or
+		 * be merged with data from `completionList.itemDefaults.data`.
+		 * 
+		 * If "replace", the data from the completion item will be used if
+		 * provided, otherwise  `completionList.itemDefaults.data` will be used.
+		 * An empty object can be used if a completion item does not have any
+		 * data but also should not use the value from
+		 * `completionList.itemDefaults.data`.
+		 * 
+		 * If "merge", a shallow merge will be performed between
+		 * `completionList.itemDefaults.data` and the completion's own data
+		 * using the following rules:
+		 * 
+		 * - If a field is specified in `completion.data` it will be used as-is.
+		 * - If a field is `null` in `completion.data`, it will remain `null`.
+		 * - If a field is unspecified in `completion.data`, the same field from
+		 *   `completionList.itemDefaults.data` will be used.
+		 *
+		 * @since 3.18.0
+		 */
+		data?: "replace" | "merge";
 	}
 
 	/**
--- a/_specifications/lsp/3.18/language/completion.md
+++ b/_specifications/lsp/3.18/language/completion.md
@@ -407,7 +407,7 @@ export interface CompletionList {
 	 * Specifies how fields from a completion item should be combined with those
 	 * from `completionList.itemDefaults`.
 	 * 
-	 * In unspecified, all fields will be treated as ApplyKind.Replace.
+	 * If unspecified, all fields will be treated as ApplyKind.Replace.
 	 * 
 	 * If a field's value is ApplyKind.Replace, the value from a completion item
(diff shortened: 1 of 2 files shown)


_specifications/lsp/3.18/language/completion.md
```diff
@@ -166,6 +166,18 @@ export interface CompletionClientCapabilities {
...
+		 * Specifies which fields of `CompletionList.applyKind` the client
+		 * supports. If omitted, no properties are supported and all fields
+		 * in a completion item will replace the defaults.
+		 * 
+		 * Clients may only specify fields that have merge rules defined in the
+		 * LSP spec.
+		 *
+		 * @since 3.18.0
+		 */
+		applyKinds?: string[];
```
rchl: This seems to rely on `itemDefaults` capability being supported. Would it make it more obvious if it would be called something like `itemDefaultsApplyKinds`?

And another discussion could be had about the "applyKinds" naming. It doesn't feel very descriptive to me (perhaps subjective). Also it refers to "kind" which is a concept that is already present in completion items that is not related to "kinds" here. Perhaps a better one would be `itemDefaultsMergeStrategy" (although it might be weird to call it "merge strategy" when the strategy can either replace or merge.

DanTup: I don't feel strongly about the name, though I think like `apply` more than `merge` since as you say, merge is one option. Maybe something like "combine" is a bit clearer? I do agree `kind` maybe could be clearer (strategy? method?)

I had a scan through the spec to see if there's anything similar we could name it the same as, but I couldn't find anything that feels the same.

_specifications/lsp/3.18/language/completion.md
```diff
@@ -386,6 +399,75 @@ export interface CompletionList {
...
+		 * `completionList.itemDefaults.data`.
+		 * 
+		 * If "merge", a shallow merge will be performed between
+		 * `completionList.itemDefaults.data` and the completion's own data
+		 * using the following rules:
+		 * 
+		 * - If a completion's data is not specified, the data from
+		 *   `completionList.itemDefaults.data` will be used.
+		 * - If a completion's data is null, the resulting data field will be
+		 *   null (defaults are not used).
```
mfussenegger: any chance this point could be removed or changed to say that if the completionItem.data is null, the itemDefaults.data is used - to treat this the same as if the field was absent? This would be a hassle to implement otherwise in neovim due to lua table semantics, and imho it would also be a bit confusing given that in pretty much all other places in the spec there is no semantic difference between a  `null` value and an absent key.

_specifications/lsp/3.18/language/completion.md
```diff
@@ -386,6 +399,75 @@ export interface CompletionList {
...
+		 * - If a field is specified in `completion.data` it will be used as-is.
+		 * - If a field is `null` in `completion.data`, it will remain `null`
+		 *   (default is not used).
```
mfussenegger: Isn't this the same point again?

_specifications/lsp/3.18/language/completion.md
DanTup: The aim here was so that there was a way to prevent using the default. So if 99.9% of your completion items have some same base data and you want to use `itemDefaults`, you can still opt-out of that default for the 0.1%.

However, in the case of `data`, maybe it's reasonable to just use an empty object (actually I think I wrote that somewhere).

Now you mention it, Dart also doesn't distinguish between `null`/`undefined` in many cases so it might also be problematic there. Lemme take another look over this.

_specifications/lsp/3.18/language/completion.md
DanTup: This one is about fields within `data`, not `data` itself:

```js
{
  label: 'MyCompletion',
  data: null,
}
```

vs

```js
{
  label: 'MyCompletion',
  data: {
    foo: null,
  },
}
```

Again, the intention was to allow `itemDefaults` to cover the majority case, but allow opting out for a subset of completion items. It's probably not critical because you could always work around this by having a boolean flag in defaults that is `true` and set to `false` on the completion item indicating to not use some other set of data.

I'll look at simplifying this too.

DanTup:
I pushed some tweaks to remove the different handling for `null` values (instead, indicating that `null` values will be treated the same as undefined). This does make it a little more difficult if you want to set some defaults and turn them off for some completion items, but I think it might not be worth the additional complexity for this.

_specifications/lsp/3.18/language/completion.md
dbaeumer: I would name it `applyKindSupport: boolean`. We should also add a `ApplyKind` definition since the spec usually doesn't use inline string or types something like

```
export namespace ApplyKind {
        export const Replace: 'replace' = 'replace;
	export const Merge: 'merge' = 'merge';
}

export type ApplyKind = ApplyKind.Replace | ApplyKind.Merge;
```

I do think one boolean is enough due to: if a client supports the new filed (e.g. via itemDefaults) and it opts into `applyKindSupport`  it MUST then support the new filed in the applyKind literal. I think that is fair.

rchl: And you are fine with using the `kind` wording even though completion items already have a [concept of `kind`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind) that is something completely different?

DanTup: @dbaeumer I've pushed changes to this affect - PTAL! I did wonder if we should use `ApplyKind` or `Completion(Item?)ApplyKind`.. I don't know if we might support replace/merge elsewhere in future and if so we'd want to use the same enum or not?

@rchl 

> And you are fine with using the kind wording even though completion items already have a [concept of kind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind)

There are a lot of `kind`s in the spec so I don't think reusing kind is bad (it's `CompletionItemKind` vs `ApplyKind`), although I'll note we also have `InsertTextMode` which is a "mode", so that's possibly another option?

_specifications/lsp/3.18/language/completion.md
```diff
@@ -525,6 +608,32 @@ export interface CompletionItemLabelDetails {
...
+	 * The value from the individual item (if provided and not `null`) will be
+	 * used instead of the default.
+	 */
+	export const Replace: 'replace' = 'replace;
+
+	/**
+	 * The value from the item will be merged with the default.
+	 * 
+	 * The specific rules for mergeing values are defined against each field
+	 * that supports merging.
```
DanTup, line 632: @dbaeumer I added this because the two existing fields are handled differently (merging of two lists vs objects), but if we think that any mergable fields will always be a list that is merged in the same way (a union) or an object that is merged in the same way (a shallow merge), then we could move those two descriptions to here and remove them from the specific fields in `CompletionList.applyKind.x`. WDYT?

dbaeumer: I would leave them on the fields since it is easier to understand how it works.

DanTup: sgtm.

If everything else here looks good to you, I can start looking at the VS Code LSP client implementation - I want to be sure we're agreed on the spec first though :)

Thanks!

dbaeumer (commented):
Looks good to me. I will merge when we have an implementation

DanTup:
Great! I've opened a PR at https://github.com/microsoft/vscode-languageserver-node/pull/1558 :-)

DanTup:
I've pushed 37f21df980f36c2dd72f23f933938bfb3b92bb7b and https://github.com/microsoft/language-server-protocol/pull/2018/commits/9081d2f979e413fd9230131534ac27d81bf642cb with tweaks discussed at https://github.com/microsoft/vscode-languageserver-node/pull/1558#discussion_r1781107923

dbaeumer (approved):

lszomoru (approved):

PR #2020, opened 2024-09-13
Support  `llmGenerated` property on `CodeAction`

StellaHuang95:
Addresses https://github.com/microsoft/vscode-languageserver-node/issues/1555

- Added `llmGeneratedSupport` to `CodeActionClientCapabilities`.
- Added `llmGenerated` to `CodeAction`.
- Marked both of them as `proposed`.


mfussenegger (commented):
Wouldn't it be more flexible if this were just a `source` hint, which can be an arbitrary string, or an enum+string for some well defined code action sources where `llm` is one of them?
E.g. if LLM gets replaced by some newer tech, there wouldn't be need to introduce a `xyGenerated` property.

Then this also wouldn't need a client capability (because if the client shows a fancy icon or something shouldn't matter for servers)

StellaHuang95:
Thanks for the feedback, I've updated to use `CodeActionTag` as suggested by Dirk. Please let me know if there's anything else you'd like me to address

dbaeumer:
@StellaHuang95 we should use the same wording tweaks as proposed in https://github.com/microsoft/vscode-languageserver-node/pull/1557

dbaeumer (changes requested):
We also need to add a client capability indicating that the client supports tags. If should go somewhere here  https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.18/language/codeAction.md#L110 and should be named `tagSupport`. You can use this https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.18/language/completion.md#L74 as a blueprint. 

Sorry, forgot about that.

dbaeumer (approved):

connor4312 (approved):

PR #2027, opened 2024-09-22
textDocumentPositionParams: add optional range

### LINKED ISSUE #377: Pass the current selection for a Hover Request
In some languages, such as Haskell, it is meaningful to show the type of an expression, rather than just an identifier.

To allow this, extend the Hover Request to also include a `Range` if the selection is active.

At the moment in the [haskell vscode plugin](https://github.com/alanz/vscode-hie-server) this is done in a roundabout way using `workspace/executeCommand`, which means it has to be separately implemented in each client.

### LINKED ISSUE #1029: Provide range or bias direction for go-to-definition etc
Go-to-definition (and others) is triggered on a cursor position. In VSCode, the cursor lies between characters, so ambiguity: the thing on the left or the thing on the right?

A couple of factors make this worse:
 - having a selection resolves the ambiguity, but the language server doesn't get to see it. In VSCode a selection usually requires biasing left.
 - other editors place the cursor *on* characters and so have no ambiguity. They always need the language server to bias right.

To illustrate, these screenshots produce the same `textDocument/definition` request, but want different results.

In vim. This should go to `Foo::operator->`.
image

In VSCode. This should go to `object`.
image

It'd be nice to have some more information so language servers can make better decisions. Some ideas:
 - requests provide optional selection range in addition to cursor position. (block-cursor editors might always provide at least a 1-char range)
 - a client capability describing the cursor type. (Language servers could implement heuristics like bias-towards-identifier for cursor-between-char editors, and have the correct behavior for block-cursor editors)

I guess this is mostly an issue with languages that have overloaded operators. constructor and functor calls are common examples in C++.

adonovan:
This change adds an optional range field to textDocumentPositionParams so that clients can, for example, query a subexpression such as x.f within a larger expression g(x.f[i]).

Also, clarify the text about cursor bias.

Fixes #377
Fixes #1029


adonovan:
@microsoft-github-policy-service agree company="Google"

dbaeumer:
@adonovan to merge something into the spec we usually require an implementation in at least one client. Are you willing to provide that as well.

And I was wondering if we at least should flag this as a feature / capability from the client side so that the servers know what to expect. In VS Code for example there will be a lot of cases where we can't fill in the range since it is not part of the original API request and getting it from the active editor is not possible (e.g. there might not be one, the cursor has been moved, ...)

adonovan:
> @adonovan to merge something into the spec we usually require an implementation in at least one client. Are you willing to provide that as well.

I already have an implementation in eglot (Emacs): it's just a few lines.

> And I was wondering if we at least should flag this as a feature / capability from the client side so that the servers know what to expect. In VS Code for example there will be a lot of cases where we can't fill in the range since it is not part of the original API request and getting it from the active editor is not possible (e.g. there might not be one, the cursor has been moved, ...)

I don't think this needs to be a capability; it's just an optional field, and its type tells the server that it can't be relied upon.

adonovan:
Sorry for not providing it earlier, but here's the diff to eglot:
```diff
 (defun eglot--TextDocumentPositionParams ()
   "Compute TextDocumentPositionParams."
   (list :textDocument (eglot--TextDocumentIdentifier)
-        :position (eglot--pos-to-lsp-position)))
+        :position (eglot--pos-to-lsp-position))   
+   (if (region-active-p)
+       (list :range 
+	     (list :start (eglot--pos-to-lsp-position (region-beginning))
+		   :end (eglot--pos-to-lsp-position (region-end)))))))
```

adonovan:
Is there anything we can do to help move this process along?

xzbdmw:
>  to merge something into the spec we usually require an implementation in at least one client

The "one client" seems always mean vscode.

MagicRB:
> > to merge something into the spec we usually require an implementation in at least one client
> 
> The "one client" seems always mean vscode.

*the* one client :) but on a serious note, an implementation exists so the "at least one implementation" requirement is fulfilled, what else is blocking this?

dbaeumer:
@MagicRB can you point me to that client implementation?

MagicRB:
> Sorry for not providing it earlier, but here's the diff to eglot:
> 
> ```diff
>  (defun eglot--TextDocumentPositionParams ()
>    "Compute TextDocumentPositionParams."
>    (list :textDocument (eglot--TextDocumentIdentifier)
> -        :position (eglot--pos-to-lsp-position)))
> +        :position (eglot--pos-to-lsp-position))   
> +   (if (region-active-p)
> +       (list :range 
> +	     (list :start (eglot--pos-to-lsp-position (region-beginning))
> +		   :end (eglot--pos-to-lsp-position (region-end)))))))
> ```

^

sideeffffect:
@tgodzik doesn't Scala's Language Server _Metals_ also use "range"?

tgodzik:
> @tgodzik doesn't Scala's Language Server _Metals_ also use "range"?

We do, we have a hack that adds range in the VS Code extension here https://github.com/scalameta/metals-vscode/blob/main/packages/metals-vscode/src/extension.ts#L411

sideeffffect:
@dbaeumer ping :cat:

PolyMeilex:
Nvim implementation: https://github.com/PolyMeilex/neovim/commit/826ff41c9cf2fd994ff68335325188aa28ed7f82

adonovan:
Hey Microsoft folks, it seems like you are all very busy on other projects. (If MS is anything like Google, I suspect AI is the reason.)

Is it time for the community to start planning how control of the protocol can be transferred to a public consortium?

nik-rev:
Helix implementation: https://github.com/helix-editor/helix/commit/1d95cc6774aa9d5b54d0a0d8e5004e8f038b5ef9

dbaeumer:
IMO we should in addition to the optional property add a client capability to signal if client in general compute that property. Otherwise it is not easy for a server to decide, whether there was no range or whether the client didn't provide it. However if we do that then clients need to consistently fill the property in. Or we can have a capability per request but that seems like overkill.

sideeffffect:
> it is not easy for a server to decide, whether there was no range or whether the client didn't provide it

I'm wondering what would be the benefit of this? What would be such a distinction useful for?

nik-rev:
> > it is not easy for a server to decide, whether there was no range or whether the client didn't provide it
> 
> I'm wondering what would be the benefit of this? What would be such a distinction useful for?

It is not always free to obtain the range - sometimes it needs to be computed. In the Helix [implementation](https://github.com/helix-editor/helix/commit/1d95cc6774aa9d5b54d0a0d8e5004e8f038b5ef9) that is the case.

Editors should not make this computation if it'll just be thrown away if the server doesn't support a range parameter. I agree it would be useful to have a capability for this

dbaeumer:
If a server wants to implement a feature that requires having that property and a client will never provided it then there is no need to register a handler for it and the client doesn't call the server unnecessarily.

charlesottesen9:
Sent from my Metro By T-Mobile 5G LTE Android Device
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

_specifications/lsp/3.18/types/textDocumentPositionParams.md
```diff
@@ -12,8 +23,14 @@ interface TextDocumentPositionParams {
 	textDocument: TextDocumentIdentifier;
 
 	/**
-	 * The position inside the text document.
+	 * The cursor position within the text document.
```
dbaeumer, line 26: Since LSP requests are also allowed for document that are not open we shouldn't talk about a cursor.

_specifications/lsp/3.18/types/textDocumentPositionParams.md
```diff
@@ -12,8 +23,14 @@ interface TextDocumentPositionParams {
...
 	 */
 	position: Position;
+
+	/**
```
dbaeumer, line 30: Same here. We need to come up with something that doesn't talk about a selection since clients might send this even for documents not open in an editor.

dbaeumer (changes requested):

PR #2032, opened 2024-10-04
Clarify snippet placeholder structure

echasnovski:
Now grammar implies that placeholder value should be a single `any` node, while in most applications it can be any number of consecutive `any` nodes. For example, as described in earlier "Placeholders" section: `${1:another ${2:placeholder}}`.

Using `any+` instead of `any` seems to be more precise.

### DIFF
--- a/_specifications/lsp/3.18/language/completion.md
+++ b/_specifications/lsp/3.18/language/completion.md
@@ -934,7 +934,7 @@ ${TM_FILENAME/(.*)\..+$/$1/}
 Below is the grammar for snippets in EBNF ([extended Backus-Naur form, XML variant](https://www.w3.org/TR/xml/#sec-notation)). With `\` (backslash), you can escape `$`, `}` and `\`. Within choice elements, the backslash also escapes comma and pipe characters. Only the characters required to be escaped can be escaped, so `$` should not be escaped within these constructs and neither `$` nor `}` should be escaped inside choice constructs.
 
 ```
-any         ::= tabstop | placeholder | choice | variable | text
+any         ::= (tabstop | placeholder | choice | variable | text)+
 tabstop     ::= '$' int | '${' int '}'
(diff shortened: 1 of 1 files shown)


ribru17:
Would it make more sense to leave these unchanged, and instead change `any` to:

```
(tabstop | placeholder | choice | variable | text)+
```

currently the grammar technically implies that there can only be one top-level `any` node in a snippet

echasnovski:
> Would it make more sense to leave these unchanged, and instead change `any` to:
> 
> ```
> (tabstop | placeholder | choice | variable | text)+
> ```
>
> currently the grammar technically implies that there can only be one top-level `any` node in a snippet

Yeah, maybe. This more feels like a style preference. I don't think grammar is a place for implications, so not sure if the first (or the broadest) definition should be treated as top level. But having an explicit mention about top level snippet parts might be useful indeed.

ribru17:
For sure. And just to be clear, either way the grammar right now is incorrect in that it only allows one `any` node at the top level (even something like `text$1` is incorrect right now because it is a sequence of two `any` nodes). I didn't want to make a PR to change `any` to `(tabstop | placeholder | choice | variable | text)+` because this one seems like it would be the perfect place to do that since it solves two issues with one. That said, we could keep this PR unchanged and then add something like 

```
source      ::= any+
```

such that now `source` is the top level node rather than any, but that seems a lot less concise

echasnovski:
I finally remembered why I chose the route of not updating `any` directly: it is a better fit for resolving #2033. Its one solution (shown in the patch) requires `any` and placeholders be separate entities. But I agree that as an isolated change modifying `any` directly should be better, so I pushed it.

@dbaeumer, maybe you have own reservations about this (and #2033, for that matter)?

ribru17:
Ah makes sense, thx for linking

PR #2087, opened 2025-01-23
Clarify that `$0` should not use any other snippet syntax

### LINKED ISSUE #1946: Clarify whether $0 can be a placeholder (`${0:foo}`)
Currently the spec says that snippets can use "tab stops and placeholders", where a tabstop looks like `$1` and a placeholder looks like `${1:foo}`.

It also says that "`$0` defines the final tab stop" but it does not make it clear whether `$0` can be used with placeholder text (like `${0:foo}`).

Using it like this works fine in VS Code (it treats it as a final tabstop and a placeholder, so the text is selected and snippet mode is exited), however nevoim's LSP client is behaving differently and it could be argued that the spec doesn't say that `$0` is allowed to be a placeholder.

@dbaeumer assuming VS Code's behaviour is desired, I can send a PR to change:

> `$0` defines the final tab stop

to

> `$0` defines the final tab stop or placeholder

Does this seem reasonable?

the-mikedavis:
This change clarifies that the zero tab stop (`$0`) should not be combined with other snippet syntax like placeholders, for example `sizeof(${0:expression-or-type})` from an older version of `clangd`. (That example should be `sizeof($0)` or `sizeof(${1:expression-or-type})$0` instead.) I believe this aligns with how VSCode treats `$0` (https://github.com/microsoft/vscode/issues/152837) as well as Neovim, and it's how we treat `$0` now in Helix (https://github.com/helix-editor/helix/pull/12647), at least for placeholders so far.

Fixes #1946

### DIFF
--- a/_specifications/lsp/3.18/language/completion.md
+++ b/_specifications/lsp/3.18/language/completion.md
@@ -879,7 +879,9 @@ The `body` of a snippet can use special constructs to control cursors and the te
 
 ##### Tab stops
 
-With tab stops, you can make the editor cursor move inside a snippet. Use `$1`, `$2`, and so on to specify cursor locations. The number is the order in which tab stops will be visited, whereas `$0` denotes the final cursor position. Multiple tab stops are linked and updated in sync.
+With tab stops, you can make the editor cursor move inside a snippet. Use `$1`, `$2`, and so on to specify cursor locations. The number is the order in which tab stops will be visited. Multiple tab stops are linked and updated in sync.
+
+The `$0` tab stop denotes the final cursor position. This tab stop should not be combined with other snippet syntax - placeholders, choices, variables or transforms - it should only take the form `$0`.
 
 ##### Placeholders
 


dbaeumer (approved):

sandy081 (approved):

DanTup:
> I believe this aligns with how VSCode treats `$0` (https://github.com/microsoft/vscode/issues/152837)

This doesn't appear to be the case, VS Code supports placeholders for $0 like others:

```ts
const doc = await vscode.workspace.openTextDocument();
const editor = await vscode.window.showTextDocument(doc);
editor.insertSnippet(new vscode.SnippetString("foo(${0:value});"));
```

image

And this is the only way to achieve good functionality when completing a function call where the user is expected to type a block of code inside it (because otherwise VS Code gets "stuck" in snippet mode and the experience is very confusing).

While I support making editors consistent, I feel like this change is bad for users/language servers because it's taking away functionality that exists and is in use today in VS Code. I don't understand why it's better to forbid including placeholder text for the final tabstop.

DanTup:
It also appears that several snippets from VS Code's own repo are using `${0}` with text:

- https://github.com/microsoft/vscode/blob/7923f036c990433adf840e22e9ba0881ab93790a/extensions/groovy/snippets/groovy.code-snippets#L223
- https://github.com/microsoft/vscode/blob/7923f036c990433adf840e22e9ba0881ab93790a/extensions/php/snippets/php.code-snippets#L55

And there's a test that seems to include this too:

https://github.com/microsoft/vscode/blob/7923f036c990433adf840e22e9ba0881ab93790a/src/vs/editor/contrib/snippet/test/browser/snippetParser.test.ts#L231

(cc @dbaeumer)

the-mikedavis:
What's the purpose of using `${0:value}` rather than `${1:value}` in that example? I'm thinking it's to avoid the behavior that `$0` is added at the end when omitted? Would `foo(${1:value}$0);` work the same?

It's news to me that this works in VSCode these days as @jrieken's comment (sorry for the ping) in the discussion I linked seemed definitive: `$0` is the editor's cue to exit the snippet mode/session. When it's combined with placeholder text you can't exit the snippet session as you need to provide the accept/replace behavior for the placeholder so it becomes unclear when the snippet mode/session is finished and whether Tab/Shift+Tab should still move between other tabstops.

DanTup:
> What's the purpose of using `${0:value}` rather than `${1:value}` in that example? I'm thinking it's to avoid the behavior that `$0` is added at the end when omitted? Would `foo(${1:value}$0);` work the same?

It wouldn't work the same, because you'd either have to press `<tab>` a final time to get to the `$0` (which is not obvious), or you'd be stick in "snippet" mode where there's a highlight over all the code you're typing and another at the end. This has been a source of many complaints in the past, for example:

- https://github.com/Dart-Code/Dart-Code/issues/1173
- https://github.com/microsoft/vscode/issues/43270

In these places, we want a final tab stop because it's where we expect the user to continue typing more code, so we don't want to be stuck in some mode that might suppress code completion, highlight everything being typed, etc. - but we also want to show some placeholder text to give the user an indication of what goes there.

I don't really understand the resistance to support `${0:some text}` (besides it not currently working in some editors and would need implementing), it is a useful feature that is in use today by multiple extensions.

HighCommander4:
> > What's the purpose of using `${0:value}` rather than `${1:value}` in that example? I'm thinking it's to avoid the behavior that `$0` is added at the end when omitted? Would `foo(${1:value}$0);` work the same?
> 
> It wouldn't work the same, because you'd either have to press `<tab>` a final time to get to the `$0` (which is not obvious), or you'd be stick in "snippet" mode where there's a highlight over all the code you're typing and another at the end.

To add to this, the desired behaviour of `foo(${0:value})` is different from `foo(${1:value}$0)` in the case of **nested** completions: if you're invoking the completion for `foo` in the context of typing a placeholder for some enclosing completion item, then with `foo(${0:value})`, after you type the "value" placeholder, a <kbd>Tab</kbd> should take you to the next placeholder of the enclosing completion. With `foo(${1:value)$0)`, the first <kbd>Tab</kbd> will be a no-op, and a second <kbd>Tab</kbd> is required to get you to the next placeholder of the enclosing completion.

Lmillan123 (approved):

dbaeumer:
I reverted the change for now to avoid any confusion.

Let's see what @jrieken has to say.

DanTup:
@dbaeumer thanks - should https://github.com/microsoft/language-server-protocol/issues/1946 be re-opened while it's being discussed?

jrieken:
In VSCode snippets I want to support TextMate snippets as good as I can and afaik in TM only `$0` is supported, not `${0:Hello}`. It is only a tabstop, not a placeholder. I think our docs/grammar mention that and this PR is correct. 

Then there is also how the VS Code implementation behaves and that might be more relaxed, e.g due to historical reasons and esp. historical bugs we might support `${0:Hello}` in common cases. Still, I don't make guarantees for this, I don't advertise this, and I don't see the need to stop doing this because it will just break existing snippets. However, new snippet authoring, guided by the docs, should use the correct form

DanTup:
> Still, I don't make guarantees for this, I don't advertise this, and I don't see the need to stop doing this because it will just break existing snippet

This stance is confusing to me. Either you plan to keep it working, or you might not. If you might not, it should IMO be deprecated and removed from the first-party snippets. That it's being used there seems evidence that it is a useful feature, and it's not clear to me who benefits from removing it.

By continuing to have it work, and use it in first-party snippets, other people will continue to use it (potentially accidentally, because it works and is being used internally). It's very difficult for extension authors, LSP server authors, and other LSP clients when the rules of what is/isn't supported are blurry.

jrieken:
> By continuing to have it work, and use it in first-party snippets,

If you see this being used in first-party snippets, then please follow-up with PRs to fix those snippets. It's your interpretation that this is a "useful feature" but I think I explained why this is accidentally supported

DanTup:
@jrieken here's some examples:

- https://github.com/microsoft/vscode/blob/7923f036c990433adf840e22e9ba0881ab93790a/extensions/groovy/snippets/groovy.code-snippets#L223
- https://github.com/microsoft/vscode/blob/7923f036c990433adf840e22e9ba0881ab93790a/extensions/php/snippets/php.code-snippets#L55

Presumably the authors of those also thought this was a useful feature. I'm not sure why placeholder text would be any less useful in the final tabstop than any other tabstop. Even if it working was accidental, forbidding it just makes the experience worse for (as far as I can tell), no actual benefit 😔

jrieken:
> @jrieken here's some examples

PR welcome

> I'm not sure why placeholder text would be any less useful in the final tabstop than any other tabstop.

Please discuss this with the authors of TextMate, not me

DanTup:
> PR welcome

If the feature will continue to be supported, I feel like me sending a PR will make them worse for no reason.


> Please discuss this with the authors of TextMate, not me

With respect, TextMate probably doesn't have the issues/quirks noted above when using a `${1}` at a location where a user is intended to write code that led to us having to use `${0}`.


@dbaeumer how do you feel about having a capability for this so clients can opt-in to this behaviour (and if it doesn't default to `true` for VS Code, we could override it in middleware while it works)?

DanTup:
I just installed TextMate and tried this there, and it works there too:

https://github.com/user-attachments/assets/a731c35b-c80d-425c-a7d8-062881296baf

DanTup:
@dbaeumer do you have an opinion on what to do here? Both VS Code and TextMate appear to support this, but it seems like there are some editors that do not. A capability would solve both, but if the intention is that TextMate is the de-facto spec, it may be unnecessary.

PR #2116, opened 2025-03-16
Fix and update tools.md

Dmole:


Dmole:

dbaeumer (changes requested):
I like to have the once separate that are basically not maintained anymore but I am not a fan of separating `Proprietary`. We don't do that anywhere else. What is the reason behind it?

Dmole:
> What is the reason behind [separating proprietary]?

This is GitHub where we share code, there is no code shared from proprietary software.
Reasons for wanting code include;

- knowing if the software is abandoned by reviewing last commit date and open issues. 
- comparing implementation details of an ambiguous or limiting API.
- figuring out how to solve a niche issue.
- fix a bug.
- find a project to contribute a pet feature to.
- security research.
- etc

dbaeumer:
I would prefer a not available indication in the repository column than a separate section.

Dmole:
Now without a 

> separate section

PR #2132, opened 2025-04-27
fix typo in glob pattern spec

brynne8:
I looked into the behavior of `*` in VS Code’s glob implementation and found that it actually matches **zero** or more characters in a path segment—not one or more as LSP 3.17 states. Specifically:

- In the VS Code source ([glob.ts](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/glob.ts#L520)), the comment reads “`*` to match zero or more characters in a path segment.”  
- The [“Glob Patterns” documentation on code.visualstudio.com](https://code.visualstudio.com/docs/editor/glob-patterns) also says “zero or more.”  
- Inspecting the implementation:

  ```ts
  function starsToRegExp(starCount: number, isLastPattern?: boolean): string {
    switch (starCount) {
      case 0:
        return '';
      case 1:
        // 1 star: match any number of characters except path separators (/ and \), non-greedy
        return `${NO_PATH_REGEX}*?`;
      default:
        // handler for **
    }
  }
  ```

  In regex syntax, `*` means “zero or more” of the preceding pattern, and the `?` makes it non-greedy.

So it looks like LSP 3.17’s description of `*` is simply a typo.

### DIFF
--- a/_specifications/lsp/3.17/metaModel/metaModel.json
+++ b/_specifications/lsp/3.17/metaModel/metaModel.json
@@ -9836,7 +9836,7 @@
 						"kind": "base",
 						"name": "string"
 					},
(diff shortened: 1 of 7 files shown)


brynne8:

dbaeumer:
Thanks. You are correct '*' should match zero or more.

dbaeumer (approved):

lszomoru (approved):

PR #2157, opened 2025-06-24
fix: `ServerCapability.referencesProvider` accepts `ReferenceRegistrationOptions` too

Sysix:
`ReferenceRegistrationOptions` is never been referenced in the specs


Sysix:

dbaeumer:
I will close the PR for the same reason as outlined in https://github.com/microsoft/language-server-protocol/pull/2158

PR #2158, opened 2025-06-24
fix: `ServerCapability.documentHighlightProvider` accepts `DocumentHighlightRegistrationOptions` too

Sysix:
`DocumentHighlightRegistrationOptions` is never been referenced in the specs

### DIFF
--- a/_specifications/lsp/3.17/general/initialize.md
+++ b/_specifications/lsp/3.17/general/initialize.md
@@ -728,7 +728,8 @@ interface ServerCapabilities {
 	/**
 	 * The server provides document highlight support.
 	 */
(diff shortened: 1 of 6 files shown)


dbaeumer:
Because of client backwards compatibility documentHighlightProvider doesn't accept DocumentHighlightRegistrationOptions. The options are not references since they can be used only for dynamic registration. See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#client_registerCapability

I will close the PR since it is from on older client perspective a breaking change.

Sysix:
Thank you! I thought it was only missing docs ❤️

PR #2159, opened 2025-06-24
fix: `ServerCapability.documentLinkProvider` accepts `DocumentLinkRegistrationOptions` too

Sysix:
`DocumentLinkRegistrationOptions` is never been referenced in the specs


dbaeumer:
@Sysix sorry for not accepting the PRs but they could break older existing clients. For more information see https://github.com/microsoft/language-server-protocol/pull/2158

Mycatty:
Ok

PR #2197, opened 2025-10-26
Use valid nested placeholder snippet example

### LINKED ISSUE #2194: Invalid snippet example in nested placeholder snippet
## Related discussions
- https://github.com/neovim/neovim/discussions/36232

`${1:another ${2:placeholder}}` (a snippet example provided [here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax)) isn't a valid snippet according to the LSP grammar.

A placeholder node is defined as `placeholder ::= '${' int ':' any '}'` where `any ::= tabstop | placeholder | choice | variable | text`. In the provided snippet the `any` (after the colon) would match another (as a text node) and then parsing crashes (as expected) because the placeholder value incorrectly consists of multiple nodes.

In order to ensure consistency across editors (as a primary goal of LSP) we need to clarify the feature descriptions and/or grammar to reflect the expected capabilities.

MariaSolOs:
Closes https://github.com/microsoft/language-server-protocol/issues/2194

I do want to call out that an alternative solution is to allow nested placeholders to consist of multiple nodes, which would require the grammar change from https://github.com/microsoft/language-server-protocol/pull/2032.

### DIFF
--- a/_specifications/lsp/3.18/language/completion.md
+++ b/_specifications/lsp/3.18/language/completion.md
@@ -883,7 +883,7 @@ With tab stops, you can make the editor cursor move inside a snippet. Use `$1`, 
 
 ##### Placeholders
 
-Placeholders are tab stops with values, like `${1:foo}`. The placeholder text will be inserted and selected such that it can be easily changed. Placeholders can be nested, like `${1:another ${2:placeholder}}`.
+Placeholders are tab stops with values, like `${1:foo}`. The placeholder text will be inserted and selected such that it can be easily changed. Placeholders can be nested, like `${1:${2:placeholder}}`.
 
 ##### Choice
 


jwortmann:
Could you explain how `${1:${2:placeholder}}` should behave in practice? A construct like that doesn't make much sense to me. Let's say I trigger such a snippet, then the text `placeholder` is inserted and selected, and when I press the <kbd>tab</kbd> key to jump to the "next" tab stop, nothing happens and it is still selected?

On the other hand I could see how something like `${1:another ${2:placeholder}}` might be useful in theory. Let's assume the full snippet is `lorem ${1:another ${2:placeholder}} ipsum`. When the snippet is triggered, the text `another placeholder` is selected first. If the user then presses <kbd>tab</kbd>, only the text `placeholder` is still selected.

Scenario 1: user accepts both placeholders: <kbd>tab</kbd><kbd>tab</kbd>
The resulting text is: `lorem another placeholder ipsum`

Scenario 2: user accepts only the first placeholder, then types something to overwrite the second placeholder: <kbd>tab</kbd><kbd>f</kbd><kbd>o</kbd><kbd>o</kbd><kbd>tab</kbd>
The resulting text is: `lorem another foo ipsum`

Scenario 3: user immediately types something to overwrite the first placeholder (that means that the second tab stop never comes into effect, because it is part of the first placeholder, which was overwritten): <kbd>f</kbd><kbd>o</kbd><kbd>o</kbd><kbd>tab</kbd>
The resulting text is: `lorem foo ipsum`

I don't know about client support for snippets with nested placeholders. But I would say, either the grammar should be fixed to allow `${1:another ${2:placeholder}}`, or nested snippets should simply not be allowed. If you instead want to add the construct `${1:${2:placeholder}}` as an explicit example for nested snippets, then the specs should include an explanation what the purpose of that is and how exactly it should behave.

*Edit:* https://github.com/microsoft/language-server-protocol/issues/2194#issuecomment-3448052770 shows a good example with a nested placeholder of the form `${1:another ${2:placeholder}}`. Since it works fine in VSCode, and I can confirm that it works in Sublime Text as well, I would vote to keep the current description and just fix the grammar accordingly.

MariaSolOs:
@jwortmann I don't have a preference over whether we should change the grammar or this example. As mentioned in the PR description https://github.com/microsoft/language-server-protocol/pull/2032 is the alternative that modifies the grammar, this PR is simply the other option.

PR #2201, opened 2025-11-20
Add inherited capabilities to JSON metaModel

PolyMeilex:
Some language bindings rely on the JSON mode to determine whether a given method is callable. It would therefore be helpful to explicitly reference the inherited capability in the JSON definition, rather than mention it only in the human-readable documentation.

### DIFF
--- a/_specifications/lsp/3.18/metaModel/metaModel.json
+++ b/_specifications/lsp/3.18/metaModel/metaModel.json
@@ -438,6 +438,8 @@
 				]
 			},
 			"messageDirection": "clientToServer",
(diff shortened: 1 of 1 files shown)


dbaeumer:
Actually the metaModel.json is generated from the TypeScript code. Instead of changing the code we would need to update the generator to emit the information.

PolyMeilex:
Is that the source of truth in question https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/protocol.ts ?

dbaeumer:
Yes, the vscode-languageserver-node repository is the truth for the meta model.

Issue #1911, opened 2024-03-21
support for generalized "references" queries


adonovan:
The LSP's `references` method takes the client editor's selected range, and returns a set of locations--of all the identifiers that refer to the same symbol as the selected identifier. A large variety of useful queries can be expressed as minor variations on this theme. For example:

- show all implicit references to the selected identifier. For example, in Go, `Point2D{1, 2}` is shorthand for `Point2D{X: 1, Y: 2}`, and it might be useful to highlight it when finding references to X or Y.
- show all [free](https://github.com/golang/tools/blob/master/gopls/doc/features/web.md#freesymbols) variables of the selected text. These are the variables that would need to become parameters in an "extract function" refactoring, but it's often useful to enumerate them even when just reading.
- show all statements that use the selected variable as an l-value. For example, `x = 1`, `x.a[i] = 2` and `p = &x` all use x as an l-value, whereas `y = x` and `print(x)` use it as an r-value. When trying to comprehend aliasing  and mutation, this difference is crucial.
- show all function declarations (including anonymous lambdas) that satisfy a particular function type (and vice versa).
- show all argument expressions that assign the selected parameter. For example, a query on y `func f(x, y int)` might report the expression `456` in the call `f(123, 456)`.
- [show](https://go.dev/cl/553255) all locations that construct a variable of the selected type (even as part of a larger aggregate).
- show all supertypes or subtypes of a selected type T. (The textDocument/implementation feature reports supertypes if T is concrete and subtypes if T is an interface, but gives no way to request superinterfaces of an interface or subtypes of a Java abstract class.)
- show all expressions that convert a value to the selected type. For example, `writer = file` might implicitly convert an `*os.File` to an `io.Writer`.
   
These examples use Go, but I'm sure you can think of other examples in your second-favorite language. ;-)

I'm not going to prescribe any particular implementation, but I think it would be very useful if the LSP allowed a server to respond to a CodeAction query with a command that tells the client: the result of this command should be displayed with a similar user interface to an ordinary 'references' query.

[...]

PR #2202, opened 2025-11-25
Queryable References

ratzdi:
A proposal for https://github.com/microsoft/language-server-protocol/issues/1911. Introduced `Reference` as new response type in the method `references`. Extended `ReferenceParams` with array of reference kinds to filter by.

### DIFF
--- a/_specifications/lsp/3.18/language/references.md
+++ b/_specifications/lsp/3.18/language/references.md
@@ -45,9 +45,71 @@ _Request_:
 <div class="anchorHolder"><a href="#referenceParams" name="referenceParams" class="linkableAnchor"></a></div>
 
 ```typescript
+export namespace ReferenceKind {
+	/**
+	 * Statement with l-value usage of the selected variable.
+	 */
+	export const Lvalue = 1;
+	/**
+	 * Statement with r-value usage of the selected variable.
+	 */
+	export const Rvalue = 2;
+	/**
+	 * Location that constructs a variable of the selected type.
+	 */
+	export const Type = 3;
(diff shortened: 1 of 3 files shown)


ratzdi:
@microsoft-github-policy-service agree [company="Dimitri Ratz"]

muradblood (approved):

muradblood (approved):

muradblood (approved):

dbaeumer:
@ratzdi besides adding something to the spec we also require a client and server implementation to prove the usefulness. Is something like this available?

ratzdi:
Hi @dbaeumer, not yet, but would it be enough if we implement the proposal on the server side in [Clangd](https://github.com/llvm/llvm-project) and on the client side in [LSP4E](https://github.com/eclipse/lsp4e)/[CDT LSP](https://github.com/eclipse-cdt/cdt-lsp)?

dbaeumer:
Yes, we don't have requirements for clients / servers.

This being said I am not a fan of the long list of reference kinds. That one is hard to maintain across different languages. A tag model would be better IMO.

qouteall:
For the `ReferenceKind` thing, I think it needs to add other reference kinds:

- For field: field initialization. I think field initialization should be different to field write.
- Import
- Re-export
- For type: usage as type parameter
- For type: type alias

And there are many other language-specific things. For example in Java a type can be used in `.class`. In Rust there are immutable borrow usages and mutable borrow usages. I think not hardcoding all reference kinds in LSP is more extensible

PR #2211, opened 2025-12-11
Fixes #2080: deprecate the top-level activeParameter property or update the documentation to clearly explain why there are two of them

### LINKED ISSUE #2080: deprecate the top-level `activeParameter` property or update the documentation to clearly explain why there are two of them
if i'm understanding the situation correctly (after skimming through https://github.com/microsoft/vscode/issues/94637), it seems that `SignatureInformation.activeParameter` is supposed to replace the top-level one in `textDocument/signatureHelp`. but currently there are two of them and [the documentation doesn't really explain why](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelp). in fact the documentation on the top level one seems to imply the opposite, since it says the following:

> In future version of the protocol this property might become mandatory to better express the active parameter if the active signature does have any.

i think either the top-level one should be explicitly marked as deprecated, or the documentation should be updated to clarify which one is preferred.

dbaeumer:
Fixes #2080

### DIFF
--- a/_specifications/lsp/3.17/language/signatureHelp.md
+++ b/_specifications/lsp/3.17/language/signatureHelp.md
@@ -221,9 +221,9 @@ export interface SignatureHelp {
 	 * lies outside the range of `signatures[activeSignature].parameters`
 	 * defaults to 0 if the active signature has parameters. If
 	 * the active signature has no parameters it is ignored.
-	 * In future version of the protocol this property might become
-	 * mandatory to better express the active parameter if the
-	 * active signature does have any.
+	 *
+	 * Since version 3.16.0 the `SignatureInformation` itself provides a
+	 * `activeParameter` property and it should be used instead of this one.
 	 */
 	activeParameter?: uinteger;
 }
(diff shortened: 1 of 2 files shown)


_specifications/lsp/3.17/language/signatureHelp.md
```diff
@@ -221,9 +221,9 @@ export interface SignatureHelp {
 	 * lies outside the range of `signatures[activeSignature].parameters`
 	 * defaults to 0 if the active signature has parameters. If
 	 * the active signature has no parameters it is ignored.
-	 * In future version of the protocol this property might become
-	 * mandatory to better express the active parameter if the
-	 * active signature does have any.
+	 *
+	 * Since version 3.16.0 the `SignatureInformation` itself provides a
+	 * `activeParameter` property and it should be used instead of this one.
```
skewb1k: Use `@deprecated` annotation? It will also allow to mark it as deprecated in the generated types

ulugbekna (approved):

PR #2213, opened 2025-12-16
Extend CallHierarchyItem with the optional array field for reference tags

ratzdi:
A reference tag indicates whether the referenced symbol is used for read or write.


ratzdi:
Closed due to incomplete PR title. Editing PR title after creation is unfortunately  not possible.

PR #2223, opened 2026-01-19
Fix typo in metaSchema

connorshea:
I could make the change to the 3.17 version as well, but I'm unsure if that is a fully locked spec that cannot be changed like this?


muradblood (approved):

dbaeumer:
@connorshea the original source for this is here: https://github.com/microsoft/vscode-languageserver-node/blob/404105914bdbbb46bec08763c21b15eb929a8ee5/tools/src/metaModel.ts#L1

I copied the changes over and the files here will update on the next release.

connorshea:
Thanks!

PR #2254, opened 2026-04-28
Add GitHub Actions workflow for npm package publishing

GitHut727:
This workflow runs tests and publishes a Node.js package to GitHub Packages upon release creation.


defenderfaith91-del:
I would give everyone the rights . But listen to me. I was hacked its a investigation going on , I had 3 computers corrupted so bad I could not get back to my data until now and its all completed by others as well as looking at what I have to agree to. Anyone who worked on the uncompleted data that was taken from my 3 computers , well I guess they decently deserve to be apart of the licence agreement hell ya, the attack is affecting my wifi and internet access , if you want  my approval on work done help me please. And I would do the correct things moving forwards.

BloodyTalonsHuisman (commented):
Approved

PR #2284, opened 2026-07-02
fix value of "Delphi"

g-plane:


dbaeumer:
This has been done on purpose. See the discussions here: https://github.com/microsoft/language-server-protocol/pull/1697

PR #2314, opened 2026-09-08
Document workspace read requests in spec

msujew:
Follow up on https://github.com/microsoft/vscode-languageserver-node/pull/1825 to update the documentation/spec.

I've made my changes against the existing 3.18 version of the spec. I hope you can use that. Let me know in case I should target a new 3.19 branch with this PR.


dbaeumer:
Since this is 3.19, I need to create a 3.19 folder for it. Will do and ping you to move it to the correct location.

